PR #2, opened 2017-08-09
linsolve: add basic types and constants

vladimir-ch:


linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// updates it, and returns the next
+	// operation. The caller must perform the
+	// Operation using data in Context, and
+	// depending on the state call Iterate
+	// again.
+	Iterate(*Context) (Operation, error)
+}
+
+// Context mediates the communication between the Method and the caller. It must
+// not be modified or accessed apart from the commanded Operations.
```
btracey: "apart from what is necessary to support the commanded Operation."

It's also someone unclear to whom this comment is addressed. I assume it's to implementers of "Method"? If so, then maybe "Methods must not modify the Context apart from ..."

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+type Context struct {
+	// X is the current approximate solution.
+	// On the first call to Method.Iterate, X
+	// must contain the initial estimate.
```
btracey: Must, or will? We will be sending this to the Method, so we control it, right?

This goes globally.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// Method must update X with the current
+	// estimate when it commands
+	// ComputeResidual and EndIteration.
+	X []float64
+	// Residual is the current residual b-A*x.
+	// On the first call to Method.Iterate,
+	// Residual must contain the initial
+	// residual.
+	// TODO(vladimir-ch): Consider whether the
```
btracey: This behavior aside, it's not clear when Residual will be changed. Will Iterate update this to pass to Context?

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package linsolve provides iterative algorithms for solving linear systems.
```
btracey: Another thing is to maybe say that the documentation is written from the perspective of users who want to call `Iterative` and designers who want to implement `Method` (assuming you agree with my comments that it should be). Users who wish to write their own version of Iterative will need to follow the same convention.

linsolve/linsolve.go
btracey: If we don't want `will`, then I think we should phrase it as "Methods may assume that ... "

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// Residual must contain the initial
+	// residual.
+	// TODO(vladimir-ch): Consider whether the
+	// behavior should also include: Method
+	// must update Residual with the current
+	// value of b-A*x when it commands
+	// EndIteration.
+	Residual []float64
+	// ResidualNorm is (an estimate of) the
+	// norm of the current residual. Method
```
btracey: Do we care which norm? The examples in the full seem to all use the 2-norm. I think it's worth saying that this is used, for example, to check convergence.

btracey: I'm not sure exactly, but there's some part of this idea that needs to be communicated to implementers of Method (who care what Context means), and users of Iterative (who don't, but do care about convergence). I'm not exactly sure yet what should be said where.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// CheckConvergence. It is not necessarily
+	// equal to the norm of Residual, some
+	// methods (e.g., GMRES) can estimate the
+	// residual norm without forming the
+	// residual itself.
+	// TODO(vladimir-ch): Actually the exact
+	// behavior here  is something that should
+	// be discussed.
+	ResidualNorm float64
+	// Converged indicates to Method that the
```
btracey: I don't understand. Can this be set either by method or by Iterative? Shouldn't Converged  = true force the Method to command EndIteration?

It also seems like we'd want to be more vague on the cause of Converged being true. My comments on the other PR aside, I suspect it's reasonable to only call an iterative method for a particular number of iterations. Then, Converged could be set to be true because the user wants to stop, not because the residual is low.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// CheckConvergence operation. If a Method
+	// commands EndIteration with Converged
+	// true, the caller must not call
+	// Method.Iterate again without calling
+	// Method.Init first.
+	Converged bool
+
+	// Src and Dst are the source and
+	// destination vectors for various
+	// Operations.
```
btracey: See the Operation documentation for more information.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	Src, Dst []float64
+}
+
+// Operation specifies the type of operation.
+type Operation uint64
```
btracey: why uint64? Isn't uint enough?

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Operations commanded by Method.Iterate.
+const (
+	NoOperation Operation = 0
+
+	// Multiply A*x where x is stored in
```
btracey: "in Context.Src, and where the result is placed in Context.Dst" / g

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// Context.Src. The result will be
+	// stored in Context.Dst.
+	MatVec Operation = 1 << (iota - 1)
```
btracey: Why MatVec and not MulVec? Maybe just `Mul` since it's the only kind of operation performed (i.e. there's no MulMat, or VecVec)

linsolve/linsolve.go
btracey: s/Multiply/Compute

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Multiply A^T*x where x is stored in
```
btracey: s/Multiply/Compute

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// stored in Context.Dst.
+	MatTransVec
+
+	// Do the preconditioner solve
+	//  M z = r,
+	// where r is stored in Context.Src, and
+	// store the solution z in Context.Dst.
+	PSolve
+
+	// Do the preconditioner solve
```
btracey: s/Do the/Perform a

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// Context.Src. The result will be
+	// stored in Context.Dst.
+	MatVec Operation = 1 << (iota - 1)
+
+	// Multiply A^T*x where x is stored in
+	// Context.Src. The result will be
+	// stored in Context.Dst.
+	MatTransVec
+
+	// Do the preconditioner solve
```
btracey: s/Do the/Perform a

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	//  M z = r,
+	// where r is stored in Context.Src, and
+	// store the solution z in Context.Dst.
+	PSolve
```
btracey: `PreconSolve`? A bit more characters, but way more clear. Maybe even just `Precon` and `PreconTrans` since that's the only kind of thing done with preconditioners.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Multiply A*x where x is stored in
+	// Context.Src. The result will be
+	// stored in Context.Dst.
+	MatVec Operation = 1 << (iota - 1)
+
+	// Multiply A^T*x where x is stored in
+	// Context.Src. The result will be
+	// stored in Context.Dst.
+	MatTransVec
```
btracey: This could then just be `MulT`. Users should be used to `T` as transpose because of `gonum/mat`

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	//  M z = r,
+	// where r is stored in Context.Src, and
+	// store the solution z in Context.Dst.
+	PSolve
+
+	// Do the preconditioner solve
+	//  M^T z = r,
+	// where r is stored in Context.Src, and
+	// store the solution z in Context.Dst.
+	PSolveTrans
```
btracey: This could then be `PreconT`

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Compute b - A*x where x is stored in
+	// Context.X and store the result into
+	// Context.Residual.
+	ComputeResidual
+
+	// Check convergence using the current
+	// approximation in Context.X and the
+	// residual norm in Context.ResidualNorm.
+	// Context.Converged must be set to
```
btracey: Again, must by who? This is done by Iterate, right?

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// indicate whether convergence has been
+	// determined, and then Method.Iterate
+	// must be called again.
+	CheckConvergence
+
+	// EndIteration indicates that Method has
+	// finished what it considers to be one
+	// iteration. It can be used to update an
```
btracey: s/can/is?

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// iteration counter. If Context.Converged
+	// is true, Context.X contains the
```
btracey: It feels like this should be "If context.Converged is true, Context.X must contain", again, since it's directed at Method implementers.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+	// approximate solution and the iterative
+	// process must be terminated. Method.Init
```
btracey: s/must be/will be?

linsolve/linsolve.go
vladimir-ch: I agree. I tried to write it from those two perspectives but I may not have been consistent enough. I will add a TODO.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package linsolve provides iterative algorithms for solving linear systems.
+package linsolve
+
+// Method is an iterative method that produces a sequence of vectors converging
```
vladimir-ch: Is this really better? It's the sequence of vectors that converges to the solution, not the type and not the method. "Method converges" is used but only as a shortcut for "the sequence of vectors that it provides converges". I will leave this as is at the moment, and come back to it later when I write the package level documentation. I may move some of this there.

linsolve/linsolve.go
```diff
@@ -0,0 +1,151 @@
...
+// to the vector x satisfying a system of linear equations
+//  A x = b,
+// where A is non-singular dim×dim matrix, and x and b are vectors of dimension
+// dim.
+//
+// Method uses a reverse-communication interface between the iterative algorithm
+// and the caller. Method acts as a client that commands the caller to perform
+// needed operations via Operation returned from the Iterate method. This
+// provides independence of Method on representation of the matrix A, and
```
vladimir-ch: And for me this is more important than the second half :-) We don't have sparse matrices in Gonum, and probably won't have for some time. I wrote this to stress that what is in this package can be used independently of that fact. If we omit the function `Iterative` there is not a single assumption on what the system matrix is. `Iterative` and `System` still keep all the freedom of the representation but they do introduce the additional `Mul` or `MulVec` function. I would like to keep this remark.

linsolve/linsolve.go
vladimir-ch: It is meant for the caller. Comment adjusted accordingly.

linsolve/linsolve.go
vladimir-ch: Here I was assuming the perspective of a direct user of `Method` where we have no control. In case of `Iterative` functions it is us. I would guess that there will be more users of Method than writers of Method. So it makes sense to keep "must" here, and a writer of `Method` can then rely on these "must" assumptions.

linsolve/linsolve.go
vladimir-ch: This is exactly the point where I'm most unsure how to define. Some methods form the residual as a part of the iteration cheaply by an axpy operation (CG, BCG). On the other hand, for example GMRES does not do this, it updates the residual only when restarting and only by forming the approximate solution first. This involves a triangular solve and matrix-vector multiply, which means it is quite costly (at least more than axpy). To make it more complated, GMRES is able to get an estimate of the _norm_ of the residual without forming it, which is the reason why I moved the responsibility for the residual norm computation into CG and BCG so that all methods behave consistently in this regard. I will ponder this a bit more.

linsolve/linsolve.go
vladimir-ch: This comes from the dilemma caused by GMRES not forming the residual as part of its process. Ideally, Method would just update Residual and the direct user (or Iterative) would use it according to their needs. But for GMRES updating Residual is expensive... As I wrote in the other comment, I will ponder this.

linsolve/linsolve.go
vladimir-ch: No, this can be set only by the caller in response to `CheckConvergence`. I don't want to specify what the Method does in response to this, it's up to the Method, it may even not respect it. Typically, it would do whatever it needs to do (again, in case of GMRES it would be forming at least X) and then command EndIteration. Simply, Method is the commander.

vladimir-ch: Regarding the second part of your comment, the caller may choose not to call `Method.Iterate` again if the maximum number of iterations has been reached after an `EndIteration` operation. The only downside is that the `Method` will not know that the caller has stopped the iteration.

linsolve/linsolve.go
vladimir-ch: I think it is. I will change it.

linsolve/linsolve.go
vladimir-ch: I wanted to use `MatVec` to highlight that this is the key matrix-vector operation. I will try how `Mul` looks in the code.

linsolve/linsolve.go
vladimir-ch: I'm starting to think that these should be written for the direct user of `Method` so I will write `... the result must be placed in Context.Dst.` Please check later and let me know what you think.

linsolve/linsolve.go
vladimir-ch: Another option would be to remove this operation and add a bool field Context.Trans. The same field could then serve `Mul` and `PSolve`.

vladimir-ch: In the end I decided to add `Context.Trans`, and go with `MulVec` and `PreconSolve`. After I commit let me know what you think.

linsolve/linsolve.go
vladimir-ch: By the caller which can be either a direct user of Method or the function `Iterative`. I rewrote the second half of this comment to clarify this.

linsolve/linsolve.go
vladimir-ch: I removed this sentence. Simply `Method` has concluded one iteration and the caller is free to use this information as it wishes.

linsolve/linsolve.go
vladimir-ch: As I wrote before, I think it is better to direct this at Method users.

vladimir-ch:
Thank you for your helpful comments, @btracey . I have now pushed the changes, so you might want to take a look again.

linsolve/linsolve.go
vladimir-ch: I've made up my mind and pushed a commits that addressed this. I changed CheckConvergence to CheckResidualNorm, and added CheckResidual to cover those methods that do and those that do not update the residual (but do get somehow it's norm).

linsolve/linsolve.go
vladimir-ch: Currently there is `MulVec` and `MulVecT`.

linsolve/linsolve.go
vladimir-ch: After some thought I decided that Context.Trans is not a good idea, and I put back the transpose operations.

linsolve/linsolve.go
vladimir-ch: Currently `PreconSolve` and `PreconSolveT`

vladimir-ch:
Motivated by your comments I've been thinking about this and made some changes. I'm happy with this and at the moment don't plan any further changes (apart from the TODO to add package documentation in the future). Please take a look again, @btracey .

linsolve/linsolve.go
```diff
@@ -0,0 +1,153 @@
...
+	// X is the current approximate solution. On the
+	// first call to Method.Iterate, X must contain the
+	// initial estimate. Method will update X with the
+	// current estimate when it commands ComputeResidual
+	// and EndIteration.
+	X []float64
+
+	// Residual is the current residual b-A*x. On the
+	// first call to Method.Iterate, Residual must
+	// contain the initial residual. Method will set it
```
btracey: I'm still kind of confused. This says "Method will set it", which to me says that the type implementing method will change the value. From what I understand, it's the Controller (or whatever we're calling Iterate) that sets it to the value, for example https://github.com/gonum/exp/pull/1/files#diff-497059e318ed3afe7d3ae1d49491db50R231

linsolve/linsolve.go
btracey: I agree that the fact that the first half is true is the most important thing. It just seemed self-evident given the fact that Method doesn't have any access to the matrix itself. I guess your point is that there is no need for wrappers or whatever (BicgstabBlockSparse).

btracey:
This reads nicely now. I think the agency is still a little confused, but I think it reads better now. If you want to submit and move on, that's fine. I imagine we'll discuss implementation changes along the way, and we can do another pass at the documentation later when the structure is worked out.

Instead of a field in `context.Trans`, we could have it be an operation `Trans`, so it would be `MulVec | Trans`

linsolve/linsolve.go
vladimir-ch: I see what you mean. It's both. If the method computes the residual as part of its process, then it will set Context.Residual without outside help. If it does not, then it will have to command ComputeResidual. This is the case of GMRES which needs it when restarting.

vladimir-ch:
I like having `Trans` operation, I will add it.

> If you want to submit and move on, that's fine.

Yes, I'd like to. After all, this is an experimental repo. I'm also still considering your comments about the alternative reverse-comm-free interface, so I'd like to experiment with that as well. As a next step I'd add `CG` so that we have a concrete implementation of `Method` to debate over.

PR #3, opened 2017-09-06
linsolve: add CG

vladimir-ch:
This PR adds the Conjugate Gradient method as a concrete implementation of `Method` together with an example on how it can be used directly without any convenience wrapper. Here this direct access to the loop is used to collect the residual norms which could later be used for example for plotting the convergence behaviour.

Real tests will come in a future PR because they need some machinery in place first.

### DIFF
--- a/linsolve/cg.go
+++ b/linsolve/cg.go
@@ -1,0 +1,100 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package linsolve
+
+import "gonum.org/v1/gonum/floats"
+
+// CG implements the Conjugate Gradient iterative method with
+// preconditioning for solving systems of linear equations
+//  A*x = b,
+// where A is a symmetric positive definite matrix.
+//
+// References:
+//  - Barrett, Richard et al. (1994). Section 2.3.1 Conjugate Gradient Method (CG).
+//    In Templates for the Solution of Linear Systems: Building Blocks for
+//    Iterative Methods (2nd ed.) (pp. 12-15). Philadelphia, PA: SIAM.
+//    Retrieved from http://www.netlib.org/templates/templates.pdf
+//  - Hestenes, M., and Stiefel, E. (1952). Methods of conjugate gradients for
+//    solving linear systems. Journal of Research of the National Bureau of
+//    Standards, 49(6), 409. doi:10.6028/jres.049.044
+//  - Málek, J. and Strakoš, Z. (2015). Preconditioning and the Conjugate Gradient
+//    Method in the Context of Solving PDEs. Philadelphia, PA: SIAM.
+type CG struct {
+	rho, rhoPrev float64
+
+	z  []float64
+	p  []float64
+	ap []float64
+
+	first  bool
+	resume int
+}
+
+// Init implements the Method interface.
+func (cg *CG) Init(dim int) {
+	if dim <= 0 {
+		panic("cg: dimension not positive")
+	}
+
+	cg.z = reuse(cg.z, dim)
+	cg.p = reuse(cg.p, dim)
+	cg.ap = reuse(cg.ap, dim)
+	cg.first = true
+	cg.resume = 1
+}
+
+// Iterate implements the Method interface. It will command the following
+// operations:
+//  MulVec
+//  PreconSolve
+//  CheckResidual
+//  EndIteration
+func (cg *CG) Iterate(ctx *Context) (Operation, error) {
+	switch cg.resume {
+	case 1:
+		ctx.Src = ctx.Residual
+		ctx.Dst = cg.z
+		cg.resume = 2
+		// Compute z_{i-1} = M^{-1} * r_{i-1}.
+		return PreconSolve, nil
+	case 2:
+		cg.rho = floats.Dot(ctx.Residual, cg.z) // ρ_{i-1} = r_{i-1} · z_{i-1}
+		if cg.first {
+			copy(cg.p, cg.z) // p_1 = z_0
+		} else {
+			beta := cg.rho / cg.rhoPrev                // β_{i-1} = ρ_{i-1} / ρ_{i-2}
+			floats.AddScaledTo(cg.p, cg.z, beta, cg.p) // p_i = z_{i-1} + β p_{i-1}
+		}
+
+		ctx.Src = cg.p
+		ctx.Dst = cg.ap
+		cg.resume = 3
+		// Compute A * p_i.
+		return MulVec, nil
+	case 3:
+		alpha := cg.rho / floats.Dot(cg.p, cg.ap)     // α_i = ρ_{i-1} / (p_i · A p_i)
+		floats.AddScaled(ctx.Residual, -alpha, cg.ap) // r_i = r_{i-1} - α A p_i
+		floats.AddScaled(ctx.X, alpha, cg.p)          // x_i = x_{i-1} + α p_i
+
+		ctx.Src = nil
+		ctx.Dst = nil
+		ctx.Converged = false
+		cg.resume = 4
+		return CheckResidual, nil
+	case 4:
+		if ctx.Converged {
+			// Calling Iterate again without Init will panic.
+			cg.resume = 0
+			return EndIteration, nil
+		}
+		cg.rhoPrev = cg.rho
+		cg.first = false
+		cg.resume = 1
+		return EndIteration, nil
+
+	default:
+		panic("cg: Init not called")
+	}
+}
--- a/linsolve/cg.go
+++ b/linsolve/cg.go
@@ -4,7 +4,9 @@
 
 package linsolve
 
-import "gonum.org/v1/gonum/floats"
+import (
+	"gonum.org/v1/gonum/floats"
+)
 
 // CG implements the Conjugate Gradient iterative method with
 // preconditioning for solving systems of linear equations
--- a/linsolve/cg_example_test.go
+++ b/linsolve/cg_example_test.go
@@ -1,0 +1,133 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package linsolve_test
+
+import (
+	"fmt"
+	"math"
+
+	"gonum.org/v1/exp/linsolve"
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/mat"
+)
+
+type system struct {
+	A *mat.SymBandDense
+	b []float64
+}
+
+// L2Projection returns a linear system whose solution is the L2 projection of f
+// into the space of piecewise linear functions defined on the given grid.
+//
+// References:
+//  - M. Larson, F. Bengzon, The Finite Element Method: Theory,
+//    Implementations, and Applications. Springer (2013), Section 1.3, also
+//    available at:
+//    http://www.springer.com/cda/content/document/cda_downloaddocument/9783642332869-c1.pdf
+func L2Projection(grid []float64, f func(float64) float64) system {
+	n := len(grid)
+
+	// Allocate the mass matrix.
+	lda := 2
+	a := make([]float64, n*lda)
+	// Assemble the mass matrix by iterating over all elements.
+	for i := 0; i < n-1; i++ {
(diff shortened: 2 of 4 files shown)


linsolve/cg.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	cg.z = reuse(cg.z, dim)
+	cg.p = reuse(cg.p, dim)
+	cg.ap = reuse(cg.ap, dim)
+	cg.first = true
+	cg.resume = 1
+}
+
+// Iterate implements the Method interface.
```
btracey: I would say something like "Iterate performs a step in converging to the solution of linear systems. See the Method interface for more information."

linsolve/cg.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"gonum.org/v1/gonum/floats"
+)
+
+// CG implements the Conjugate Gradient iterative method with
+// preconditioning for solving systems of linear equations
+//  A*x = b,
+// where A is a symmetric positive definite matrix.
+//
+// CG will command the following operations:
```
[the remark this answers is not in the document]
btracey: Could you include some kind of reference?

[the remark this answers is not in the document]
vladimir-ch: I think that having the commanded Operations documented somewhere is necessary, otherwise someone using Init/Iterate directly would have to either handle all Operations or rely on examples. I will move this to Iterate.

I don't have a reference specifically for CG. All the good references I have are about iterative methods in general, I will add it to the package-level documentation.

linsolve/cg.go
vladimir-ch: Then I would have to repeat this with every iterative method. I will see if I can add this sentence to Method.Iterate.

linsolve/cg.go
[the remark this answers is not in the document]
vladimir-ch: Ok, I added the original CG article which is very nice despite its age. I also added a recent book on CG which is quite advanced but very nice and presents a modern approach to the theory (and practice too).

linsolve/cg.go
vladimir-ch: Done.

vladimir-ch:
PTAL

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+	ap []float64
+}
+
+// Init implements the Method interface.
+func (cg *CG) Init(dim int) {
+	if dim <= 0 {
+		panic("cg: dimension not positive")
+	}
+
+	cg.z = reuse(cg.z, dim)
```
btracey: Could you reorganize these in the same order of the struct?

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+//  CheckResidual
+//  EndIteration
+func (cg *CG) Iterate(ctx *Context) (Operation, error) {
+	switch cg.resume {
+	case 1:
+		ctx.Src = ctx.Residual
+		ctx.Dst = cg.z
+		cg.resume = 2
+		return PreconSolve, nil
+	case 2: // Compue z_{i-1} = M^{-1} * r_{i-1}.
```
btracey: Isn't this comment for case 1? `z_{i-1} = M^{-1} * r_{i-1}` is a precon solve.

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+		cg.rho = floats.Dot(ctx.Residual, cg.z) // ρ_{i-1} = r_{i-1} · z_{i-1}
```
btracey: I'm not sure, but this may be more clear as `ctx.Dst`.

The high level comment is that it's a little tricky that the data is shared between `ctx` and `Method`. I can see the purpose, and I imagine the performance differences are enough to make it worth it, but it's a bit different than most of the gonum APIs. I wonder if there's an easy way to comment somewhere that the Method is responsible for allocating data for `Context`, instead of it happening in the high level code (as it does in `optimize`)

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+		} else {
+			beta := cg.rho / cg.rhoPrev                // β_{i-1} = ρ_{i-1} / ρ_{i-2}
+			floats.AddScaledTo(cg.p, cg.z, beta, cg.p) // p_i = z_{i-1} + β p_{i-1}
+		}
+
+		ctx.Src = cg.p
+		ctx.Dst = cg.ap
+		cg.resume = 3
+		return MulVec, nil
+	case 3: // Compute A*p_i.
```
btracey: Same about this comment, seems like it refers to the above line

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+		alpha := cg.rho / floats.Dot(cg.p, cg.ap)     // α_i = ρ_{i-1} / (p_i · A p_i)
+		floats.AddScaled(ctx.Residual, -alpha, cg.ap) // r_i = r_{i-1} - α A p_i
+		floats.AddScaled(ctx.X, alpha, cg.p)          // x_i = x_{i-1} + α p_i
+
+		ctx.Src = nil
+		ctx.Dst = nil
+		ctx.Converged = false
+		cg.resume = 4
+		return CheckResidual, nil
```
btracey: I think this should be called `CheckConvergence` or something that conveys it's a "maybe stop" and not that the caller has to compute `A*x-b`.

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+		ctx.Src = cg.p
+		ctx.Dst = cg.ap
+		cg.resume = 3
+		return MulVec, nil
+	case 3: // Compute A*p_i.
+		alpha := cg.rho / floats.Dot(cg.p, cg.ap)     // α_i = ρ_{i-1} / (p_i · A p_i)
+		floats.AddScaled(ctx.Residual, -alpha, cg.ap) // r_i = r_{i-1} - α A p_i
+		floats.AddScaled(ctx.X, alpha, cg.p)          // x_i = x_{i-1} + α p_i
+
+		ctx.Src = nil
```
btracey: Is this just a defensive nil setting?

linsolve/cg_example_test.go
```diff
@@ -0,0 +1,133 @@
...
+	h := (x1 - x0) / float64(n)
+	grid := make([]float64, n+1)
+	for i := range grid {
+		grid[i] = x0 + float64(i)*h
+	}
+	grid[n] = x1
+	return grid
+}
+
+func ExampleCG() {
```
btracey: Can we name this something less general? I assume that in the end we will have a high-level function we want users to call rather than writing their own loop. It's good to have an example without the high-level call, but I don't think it should be the default example for CG

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+package linsolve
+
+import (
+	"gonum.org/v1/gonum/floats"
+)
+
+// CG implements the Conjugate Gradient iterative method with
+// preconditioning for solving systems of linear equations
+//  A*x = b,
+// where A is a symmetric positive definite matrix.
```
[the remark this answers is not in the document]
vladimir-ch: I think we could mention it at the package level. Sparse BLAS supports matrix properties in case we provide it someday but these annotations come from the user anyway.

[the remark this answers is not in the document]
btracey: Yea, I think the package level is fine. It's kind of obvious, but we design APIs to prevent this kind of mistake with Dense matrices, so I think it's worth mentioning that we can't help them here

linsolve/cg.go
vladimir-ch: Yes because Method "owns" `Src` and `Dst`.

linsolve/cg.go
vladimir-ch: There is also `Context.ResidualNorm` so I want to distinguish between checking convergence based on this and ResidualNorm. To compute `A*x-b` there is `ComputeResidual`.

linsolve/cg.go
```diff
@@ -0,0 +1,96 @@
...
+//
+// References:
+//  - Hestenes, M., and Stiefel, E. (1952). Methods of conjugate gradients for
+//    solving linear systems. Journal of Research of the National Bureau of
+//    Standards, 49(6), 409. doi:10.6028/jres.049.044
+//  - Málek, J. and Strakoš, Z. (2015). Preconditioning and the Conjugate Gradient
+//    Method in the Context of Solving PDEs. Philadelphia, PA: SIAM.
```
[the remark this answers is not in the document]
vladimir-ch: Done.

linsolve/cg.go
vladimir-ch: I reorganized the struct, I want `resume` to be the last before returning.

linsolve/cg.go
```diff
@@ -56,7 +60,8 @@ func (cg *CG) Iterate(ctx *Context) (Operation, error) {
 		ctx.Dst = cg.z
 		cg.resume = 2
 		return PreconSolve, nil
-	case 2: // Compue z_{i-1} = M^{-1} * r_{i-1}.
+		// Compute z_{i-1} = M^{-1} * r_{i-1}.
```
btracey: I think this would be better above the return (it's what PreconSolve does)

linsolve/cg.go
```diff
@@ -69,7 +74,8 @@ func (cg *CG) Iterate(ctx *Context) (Operation, error) {
 		ctx.Dst = cg.ap
 		cg.resume = 3
 		return MulVec, nil
-	case 3: // Compute A*p_i.
+		// Compute A*p_i.
```
btracey: Same thing, I would put it above the return

linsolve/doc.go
```diff
@@ -2,7 +2,19 @@
...
+   Retrieved from http://www.netlib.org/templates/templates.pdf
+ - Saad, Yousef (2003). Iterative methods for sparse linear systems (2nd ed.).
+   Philadelphia, PA: SIAM.
+   Retrieved from http://www-users.cs.umn.edu/~saad/IterMethBook_2ndEd.pdf
+ - Greenbaum, A. (1997). Iterative methods for solving linear systems.
+   Philadelphia, PA: SIAM.
+*/
 package linsolve
 
 // TODO(vladimir-ch): Improve documentation. Write an introduction about
```
btracey: Could you add a TODO for specifying how the data in Context works (who needs to create what). We'll have a better sense of what is actually needed when we're farther along.

linsolve/cg.go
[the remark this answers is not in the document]
vladimir-ch: Added to the TODO list.

linsolve/cg.go
vladimir-ch: I put it here to document that it is at this point where the caller does the operation. I now see that it is confusing, so I moved it before the return.

linsolve/cg.go
vladimir-ch: I don't understand the first part of the comment.

Regarding the second part: I think it is not much different from `optimize`, only here Method needs the caller to do a larger variety of things compared to `optimize` and to do them on data that is internal to Method. Like in `optimize`, the high-level code allocates Context.X and Context.Residual. Method allocates its internal slices and temporarily assigns them to Context.Src and Context.Dst. The caller must not keep them. This is a point that should be documented.

Originally I thought about Method.Init returning the size of a work slice that the caller should allocate but then it would be cumbersome for Method to specify which part of the work slice is source and which is destination. The current approach with Src and Dst is much simpler.

linsolve/cg_example_test.go
vladimir-ch: At the moment I will leave it as `ExampleCG`. Later when I add the high-level function, there will definitely be an example and it will be easier to come up with a better name.

linsolve/doc.go
vladimir-ch: Done.

linsolve/cg.go
btracey: In Optimize, the `Context` struct would be constructed with `Src` and `Dst` having a size of `Dim`. Then, rather than `Method` allocating `cg.z`, instead it would just use `ctx.Dst` directly. Similarly with `cg.p` being set as the `Src`.

I think that actually may be a better mechanism. We already say that the caller can't modify fields of `Context` without specific permission. We can also say that `Method` may assume that `Context.Src` and `Context.Dst` have size `dim`. I think this would be a lot more clear than the shadowing going on here.

I'd be happy to make a PR showing these changes if that's confusing.

PR #4, opened 2017-09-15
Ctxdst

btracey:


btracey:
@vladimir-ch this has the memory more in line with the other packages (IMO)

btracey:
If it turns out copying is the bottleneck, we can also say the Method has to set Src.

vladimir-ch:
I think that almost always the bottleneck will be the matrix-vector multiply, I just wanted to try not to add to it. I like how it is now.

PR #6, opened 2017-09-23
linsolve: add helpers for testing and test for CG

vladimir-ch:


linsolve/cg_test.go
```diff
@@ -0,0 +1,101 @@
...
+		randomSPD(200, rnd),
+		randomSPD(500, rnd),
+		market("nos1", 1e-12),
+		market("nos4", 1e-12),
+		market("nos5", 1e-12),
+		market("bcsstm20", 1e-12),
+		market("bcsstm22", 1e-11),
+	} {
+		n := tc.n
+		// Compute the right-hand side b so that the vector [1,1,...,1]
```
btracey: Could we have the sentinel vector not be all 1s? That seems too likely to be an initial condition or particular focal point of an iterative algorithm. Maybe `want[i] = 2 + 0.1*float64(i)`?

linsolve/cg_test.go
```diff
@@ -0,0 +1,101 @@
...
+			X:        make([]float64, n),
+			Residual: make([]float64, n),
+			Src:      make([]float64, n),
+			Dst:      make([]float64, n),
+		}
+		// The initial guess is the zero vector [0,0,...,0], so the initial
+		// residual b-A*x is just b.
+		copy(ctx.Residual, b)
+
+		if floats.Norm(ctx.Residual, 2) < tc.tol {
```
btracey: I'm not sure what the right resolution is, but this condition is effectively a non-test. It seems like that this should be noisy somehow so the test can be re-arranged or fixed.

linsolve/internal/mmarket/reader.go
```diff
@@ -0,0 +1,103 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mmarket
```
btracey: This needs a comment about what it is and what it does

linsolve/internal/mmarket/reader.go
```diff
@@ -0,0 +1,103 @@
...
+	s *bufio.Scanner
+}
+
+func NewReader(r io.Reader) *Reader {
+	return &Reader{
+		s: bufio.NewScanner(r),
+	}
+}
+
+func (r *Reader) Read() (*triplet.Matrix, error) {
```
btracey: Comment here too.

linsolve/internal/triplet/triplet.go
```diff
@@ -0,0 +1,58 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package triplet
```
btracey: Comment here also.

linsolve/internal/triplet/triplet.go
```diff
@@ -0,0 +1,58 @@
...
+
+type triplet struct {
```
btracey: Not to be addressed here, but now that we are developing code for sparse matrices, it might be worth developing a list of types we might want to have in a `gonum/mat/sparse` and what methods we want those implementations to have. I'm not sure if we want some kind of tracking issue?

linsolve/internal/triplet/triplet.go
```diff
@@ -0,0 +1,58 @@
...
+	i, j int
+	v    float64
+}
+
+type Matrix struct {
+	r, c  int
+	data  []triplet
```
btracey: Make this a map from struct{r,c int} to float64?

linsolve/internal/triplet/triplet.go
```diff
@@ -0,0 +1,58 @@
...
+		r: r,
+		c: c,
+	}
+}
+
+func (m *Matrix) Dims() (r, c int) {
+	return m.r, m.c
+}
+
+func (m *Matrix) Append(i, j int, v float64) {
```
btracey: Just `Set`? This needs a comment that it doesn't check the element has been set, or just use the map if you take the above suggestion.

linsolve/internal/triplet/triplet.go
```diff
@@ -0,0 +1,58 @@
...
+	if i < 0 || m.r <= i {
+		panic("triplet: row index out of range")
+	}
+	if j < 0 || m.c <= j {
+		panic("triplet: column index out of range")
+	}
+	m.data = append(m.data, triplet{i, j, v})
+}
+
+func (m *Matrix) MulVec(dst, x []float64, trans bool) {
```
btracey: Thinking about it, I no longer like `MulVec` as the name of this method. It's fine in the context of the package, but it's confusing in the context of `*Dense`, where vector means `mat.Vector`. I wonder if there's an easy way to fix this confusion. It would be nice if we could make `*Dense` implement these so they could be passed directly.

linsolve/linsolve_test.go
```diff
@@ -0,0 +1,70 @@
...
+	iters  int
+	tol    float64
+	mulvec func(dst, x []float64, trans bool)
+}
+
+// randomSPD returns a random symmetric positive-definite matrix of order n.
+func randomSPD(n int, rnd *rand.Rand) testCase {
+	a := make([]float64, n*n)
+	lda := n
+	for i := 0; i < n; i++ {
```
btracey: You should use `*SymDense.SymOuterK`, (plus probably a diagonal regularizer) to guarantee PSD.

linsolve/linsolve_test.go
```diff
@@ -0,0 +1,70 @@
...
+		a[i*lda+i] += float64(n)
+	}
+	return testCase{
+		name:  "randomSPD",
+		n:     n,
+		iters: 2 * n,
+		mulvec: func(dst, x []float64, _ bool) {
+			bi := blas64.Implementation()
+			bi.Dsymv(blas.Upper, n, 1, a, lda, x, 1, 0, dst, 1)
```
btracey: Use the `mat` functions, we don't need BLAS here.

linsolve/cg_test.go
vladimir-ch: What do you mean by "sentinel vector"? The solution? The right-hand side is not given, so I'm free to chose here the solution to something convenient. It is easier to see which components are wrong in a vector that's correctly all 1's rather than in a random vector.

On the other hand, it can be more challenging to use a random vector for the initial guess, so I changed the code to use that rather than the zero vector.

linsolve/cg_test.go
vladimir-ch: I was aware of this being a non-test. It was a kind of safety check in case the test (the initial guess) is changed in the future to some unexpected value. I have removed this check.

linsolve/internal/mmarket/reader.go
vladimir-ch: Done.

linsolve/internal/mmarket/reader.go
vladimir-ch: Done.

linsolve/internal/triplet/triplet.go
vladimir-ch: Done.

linsolve/internal/triplet/triplet.go
vladimir-ch: I will add an issue for this.

linsolve/internal/triplet/triplet.go
vladimir-ch: I had this but the unspecified iteration order makes it unsuitable for testing.

linsolve/internal/triplet/triplet.go
vladimir-ch: If it were `Set`, then it should check whether the element exists. With the triplet representation this would be impractical, so `Append` (or `Insert`) is a more accurate name in my opinion.

linsolve/internal/triplet/triplet.go
vladimir-ch: So far the discussion has been about the name of `Operation`, not about a method. Also this is why I wanted to avoid depending on any matrix types or interfaces in `linsolve`, and why I have `triplet` as an internal package. I agree that this name could be confusing in the context of `mat` and that it would be nice to eventually have `*Dense` and the sparse types satisfy some common interface for this purpose, but that is a separate discussion.

linsolve/linsolve_test.go
vladimir-ch: Ok, done.

vladimir-ch:
PTAL

linsolve/cg_test.go
btracey: I meant more the "golden vector", i.e. the answer.  I wasn't suggesting a random vector, just a known vector that is less special, for example, [2.0, 2.1, 2.2, 2.3, ... ]. It's easy to see which element(s) are wrong, but it's less likely to be accidentally correct.

linsolve/internal/triplet/triplet.go
btracey: Sorry, you're right. I was confusing with what I assume will eventually be the `System` interface (spent too much time looking at the full PR I guess)

linsolve/internal/triplet/triplet.go
btracey: Could you add a comment to append says that it adds a non-zero element and does not check for existance

PR #7, opened 2017-10-20
linsolve: add GMRES

vladimir-ch:


linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+	// and a better, problem-dependent value of
+	// Restart should be used instead.
+	Restart int
+
+	s  []float64
+	y  []float64
+	av []float64
+
+	// v is an n×(m+1) matrix (where m == Restart)
+	// in column-major storage. Its columns form
```
vladimir-ch: This has been my main dilemma. Mathematics stores vectors in the columns of a matrix. All GMRES descriptions follow this. I'd like to avoid strided copies between columns of V and Context.Src and Dst. Alternatively, we can say that we are storing the vectors in rows and that our V and H are V^T and H^T in a standard description of GMRES.

btracey: I think that's probably better than the column major storage. I would call the fields `Vt` and `Ht` and comment that we are storing the transpose so that vectors are rows.

linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/floats"
+)
+
+// GMRES implements the Generalized Minimum Residual method with the modified
```
btracey: s/with the modified/with modified

linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+// Gram-Schmidt orthogonalization  for solving systems of linear equations
```
btracey: remove extra space

linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+//  A*x = b,
+// where A is a square, non-symmetric matrix. It uses restarts to control
+// storage requirements.
+//
+// References:
+//  - Barrett, R. et al. (1994). Section 2.3.4 Generalized Minimal Residual (GMRES).
+//    In Templates for the Solution of Linear Systems: Building Blocks for
+//    Iterative Methods (2nd ed.) (pp. 12-15). Philadelphia, PA: SIAM.
```
btracey: This says p12-15, but the actual GMRES algorithm is on page 18

linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+//    Retrieved from http://www.netlib.org/templates/templates.pdf
+//  - Saad, Y., and Schultz, M. (1986). GMRES: A generalized minimal residual
+//    algorithm for solving nonsymmetric linear systems. SIAM J. Sci. Stat. Comput., 7(3), 856.
+//    doi:10.6028/jres.049.044
+//    Retrieved from https://web.stanford.edu/class/cme324/saad-schultz.pdf
+type GMRES struct {
+	// Restart is the restart parameter which
+	// limits the computation and storage costs.
+	// It must hold that
+	//  0 <= Restart <= n
```
btracey: Is it the case that really 1 <= Restart <= n? I think that's more clear for the comment if so (a valid value is in this range, but the default value gives you n).

linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+	// where n is the dimension of the problem.
+	// If it is 0, it will be set to n. Note
```
btracey: I don't like the way this comment reads. It sounds like it says "The default value of the parameter is a bad idea"

I'm not an expert in GMRES, but it sounds like a better default is hard to choose because of convergence guarantees. I would instead say something like `If Restart is 0, it will be set to n. This guarantees convergence of GMRES and increases robustness. Many specific problems however, particularly for large n, will benefit in efficiency by limiting the memory to a problem-dependent value less than n`

linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+//  - Barrett, R. et al. (1994). Section 2.3.4 Generalized Minimal Residual (GMRES).
+//    In Templates for the Solution of Linear Systems: Building Blocks for
+//    Iterative Methods (2nd ed.) (pp. 12-15). Philadelphia, PA: SIAM.
+//    Retrieved from http://www.netlib.org/templates/templates.pdf
+//  - Saad, Y., and Schultz, M. (1986). GMRES: A generalized minimal residual
+//    algorithm for solving nonsymmetric linear systems. SIAM J. Sci. Stat. Comput., 7(3), 856.
+//    doi:10.6028/jres.049.044
+//    Retrieved from https://web.stanford.edu/class/cme324/saad-schultz.pdf
+type GMRES struct {
+	// Restart is the restart parameter which
```
btracey: Widen these comments out to 80 character lines.

linsolve/gmres.go
```diff
@@ -0,0 +1,288 @@
...
+		hjk := floats.Dot(vj, w)
+		hk[j] = hjk                   // H[j,k] = v_j · v_{k+1}
+		floats.AddScaled(w, -hjk, vj) // v_{k+1} -= H[j,k] * v_j
+	}
+	wnorm := floats.Norm(w, 2)
+	hk[k+1] = wnorm          // H[k+1,k] = |v_{k+1}|
+	floats.Scale(1/wnorm, w) // Normalize v_{k+1}.
+}
+
+// qr uses Givens rotations to reduce H to upper-triangular form.
```
btracey: Is there something special about GMRES where we can't use the Lapack equivalent of this? For instance, Dgehrd?

btracey:
Higher level comment is that it would be best to use `mat` rather than `blas` directly. If we need to add to `mat` to make this possible, let's see if we can.

linsolve/gmres.go
vladimir-ch: I kept `with the modified Gram-Schmidt`, I think it is actually an established name for the algorithm. Let me know if my justification is wrong (outside of textbook examples the rules for using "the", "a" or nothing still seem rather esoteric to me).

linsolve/gmres.go
vladimir-ch: Done. Thanks.

linsolve/gmres.go
vladimir-ch: Thanks, fixed.

linsolve/gmres.go
vladimir-ch: Done although I'm aware that @kortschak prefers field comments narrow.

linsolve/gmres.go
vladimir-ch: Yes, `1 <= Restart <= n` is correct but I didn't want to exlude 0 if it's actually allowed as a placeholder for a default value.

linsolve/gmres.go
vladimir-ch: I changed to what you suggest (although the part about a bad idea is in fact true for large n).

linsolve/gmres.go
vladimir-ch: `Dgehrd` probably not, it works on a matrix as a whole while GMRES builds the factorization one column at each iteration. `Dlarf*` functions could be useful if we used the Householder reflectors to build the factorization. This is in fact an existing variant of GMRES that may be more stable but is also more expensive. I would like to implement it but it would probably deserve its own type.

vladimir-ch:
PTAL

bq. Higher level comment is that it would be best to use mat rather than blas directly.

Yeah, probably it's because I feel more comfortable with BLAS. If I were to use it, I think the only missing piece would be getting a view of `Dense` as a triangular matrix. I'm not sure if it makes sense adding something like it to `mat`, it looks quite special. Passing it then to `SolveVec` should already work. The rest could be done with `RawRowView`.

linsolve/gmres.go
btracey: If it's an established name, that's fine. 

Okay, so I wrote half of a paragraph, and realized I can't come up with the rule on "the" vs. no "the". The fundamental tension, I believe, is the multiple meanings of the word "with", and whether it means "using" or whether it means "that has", in other words, whether what comes after the "with" modifies the noun or the verb. So, we could say that "package linsolve solves systems of linear equations with an iterative method". Here "with" is a synonym for "using", it is how we are solving the system. On the other hand, we may have "Halton implements the Halton sequence with scrambling". In this case, it's not saying how we are implementing the Halton sequence, it's saying we're implementing a specific kind of Halton sequence (that uses scrambling). So, in the case here, it reads to me as "using the Gram-Schmidt", whereas I think it means "the kind of GMRES that has modified Gram-Schmidt".

On the other hand, I can think of obvious counterexamples to my argument above, so it's certainly not a strong rule. English is dumb.

kortschak: Articles are the greatest shibboleth.

btracey: And prepositions

linsolve/gmres.go
kortschak: The issue to consider here is how it is rendered in godoc, since we can't rely on godoc to do text flow for us with struct comments. In code I prefer narrow comments to ease reading effort (comments there are often more detailed or subtle and so this helps), these comments look OK at the longer length I think, but consider how it renders on a smaller screen via godoc.org.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+		g.resume = 5
+		// Solve M^{-1} * (A * v_k).
+		return PreconSolve, nil
+	case 5:
+		// v_{k+1} = M^{-1} * (A * v_k)
+		copy(g.vcol(g.k+1), ctx.Dst)
+		// Construct (k+1)-th column of the upper Hessenberg matrix H using
+		// the modified Gram-Schmidt process on V and v_{k+1} so that it is
+		// orthonormal to the previous k columns.
+		g.modifiedGS()
```
btracey: `g.modifiedGS(g.k, g.vt, g.hcol(k), ctx.Dst)`   See comment below.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+		// Compute A * v_k.
+		return MulVec, nil
+	case 4:
+		copy(ctx.Src, ctx.Dst)
+		g.resume = 5
+		// Solve M^{-1} * (A * v_k).
+		return PreconSolve, nil
+	case 5:
+		// v_{k+1} = M^{-1} * (A * v_k)
+		copy(g.vcol(g.k+1), ctx.Dst)
```
btracey: I think this copy is harmful for following the pseudocode in the templates. I'd rather the copy happen in `modifiedGS`

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+		// Construct (k+1)-th column of the upper Hessenberg matrix H using
+		// the modified Gram-Schmidt process on V and v_{k+1} so that it is
```
btracey: This is supposed to mean `and also compute v_{k+1} ...`, yes?

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+		return EndIteration, nil
+
+	default:
+		panic("gmres: Init not called")
+	}
+}
+
+// modifiedGS applies the modified Gram-Schmidt procedure on the (k+1)-th row of V^T.
+func (g *GMRES) modifiedGS() {
```
btracey: It's good to have `modifiedGS` be a method for namespace reasons, but this is effectively a function to operate on some data, rather than really needing an object-oriented call. As such, I'd much rather this take in as arguments what it needs, and not reference the data in `GMRES` at all.

```
func (*GMRES) modifiedGS(k int, vt []float64, hk []float64, w []float64)
```

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+		hk[j] = hjk                   // H[j,k] = v_j · v_{k+1}
+		floats.AddScaled(w, -hjk, vj) // v_{k+1} -= H[j,k] * v_j
+	}
+	wnorm := floats.Norm(w, 2)
+	hk[k+1] = wnorm          // H[k+1,k] = |v_{k+1}|
+	floats.Scale(1/wnorm, w) // Normalize v_{k+1}.
+}
+
+// qr uses Givens rotations to reduce H to upper-triangular form.
+func (g *GMRES) qr() {
```
btracey: Same comment as above for using input arguments.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+	hk := g.hcol(k)
+	for j := 0; j <= k; j++ {
+		vj := g.vcol(j)
+		hjk := floats.Dot(vj, w)
+		hk[j] = hjk                   // H[j,k] = v_j · v_{k+1}
+		floats.AddScaled(w, -hjk, vj) // v_{k+1} -= H[j,k] * v_j
+	}
+	wnorm := floats.Norm(w, 2)
+	hk[k+1] = wnorm          // H[k+1,k] = |v_{k+1}|
+	floats.Scale(1/wnorm, w) // Normalize v_{k+1}.
```
btracey: Then put the copy here  (in replace of before the call).

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+	m := g.Restart
+	if m <= 0 || dim < m {
+		panic("gmres: invalid value of Restart")
+	}
+
+	g.s = reuse(g.s, m+1)
+	g.y = reuse(g.y, dim)
+	g.av = reuse(g.av, dim)
+
+	ldv := dim
```
btracey: Do these also need to be zeroed? Looks like it.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+	bi.Dtrsv(blas.Lower, blas.Trans, blas.NonUnit, k, g.ht, ldh, y, 1)
+
+	// Update the current solution vector x.
+	for j, yj := range y {
+		vj := g.vcol(j)
+		floats.AddScaled(x, yj, vj) // x += y_j * v_j
+	}
+}
+
+// drotg returns Givens plane rotation.
```
btracey: Use the BLAS function?

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+		tmp := -a / b
+		s := 1 / math.Sqrt(1+tmp*tmp)
+		return givens{c: tmp * s, s: s}
+	}
+	tmp := -b / a
+	c := 1 / math.Sqrt(1+tmp*tmp)
+	return givens{c: c, s: tmp * c}
+}
+
+// drot applies Givens rotation g to the vector [x,y] and returns the result.
```
btracey: Use the BLAS function?

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+		return PreconSolve, nil
+	case 5:
+		// v_{k+1} = M^{-1} * (A * v_k)
+		copy(g.vcol(g.k+1), ctx.Dst)
+		// Construct (k+1)-th column of the upper Hessenberg matrix H using
+		// the modified Gram-Schmidt process on V and v_{k+1} so that it is
+		// orthonormal to the previous k columns.
+		g.modifiedGS()
+		// Reduce H back to upper triangular form.
+		g.qr()
```
btracey: Is there a better description of the `J_1, ...`. It looks like what you're doing here, just curious if there's an easier way to check than me thinking through it.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+
+		ctx.ResidualNorm = g.residualNorm()
```
btracey: Inputs here too I think.

linsolve/gmres_test.go
```diff
@@ -0,0 +1,118 @@
...
+		randomSPD(2, rnd),
+		randomSPD(3, rnd),
+		randomSPD(4, rnd),
+		randomSPD(5, rnd),
+		randomSPD(10, rnd),
+		randomSPD(20, rnd),
+		randomSPD(50, rnd),
+		randomSPD(100, rnd),
+		randomSPD(200, rnd),
+		randomSPD(500, rnd),
```
btracey: Other random test matrices too since this works for more than SPD?

linsolve/gmres.go
vladimir-ch: If I do it in `modifiedGS`, then it does too many things and documenting it clearly is next to impossible. I really tried but I think it is better this way.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+	if cap(g.givs) < m {
+		g.givs = make([]givens, m)
+	} else {
+		g.givs = g.givs[:m]
+		for i := range g.givs {
+			g.givs[i].c = 0
+			g.givs[i].s = 0
+		}
+	}
+
```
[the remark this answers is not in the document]
vladimir-ch: I thought about it but it is already correctly reset before the inner iteration starts, so resetting it also here would be strange.

linsolve/gmres.go
vladimir-ch: The BLAS operates on two vectors, not on a 2d vector. I renamed it to `applyGivens`

linsolve/gmres.go
vladimir-ch: Done.

linsolve/gmres.go
vladimir-ch: Done.

linsolve/gmres.go
vladimir-ch: Done.

linsolve/gmres.go
vladimir-ch: Changed the code here a bit.

linsolve/gmres.go
vladimir-ch: Reworked the comment after changing `modifiedGS`.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+//  ComputeResidual
+//  CheckResidual
+//  CheckResidualNorm
+//  EndIteration
+func (g *GMRES) Init(dim int) {
+	if dim <= 0 {
+		panic("gmres: dimension not positive")
+	}
+
+	if g.Restart == 0 {
```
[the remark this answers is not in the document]
vladimir-ch: Done.

linsolve/gmres.go
```diff
@@ -0,0 +1,287 @@
...
+
+	"gonum.org/v1/gonum/blas"
+	"gonum.org/v1/gonum/blas/blas64"
+	"gonum.org/v1/gonum/floats"
+)
+
+// GMRES implements the Generalized Minimum Residual method with the modified
+// Gram-Schmidt orthogonalization for solving systems of linear equations
+//  A*x = b,
+// where A is a square, non-symmetric matrix. It uses restarts to control
```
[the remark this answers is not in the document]
vladimir-ch: Done.

linsolve/gmres_test.go
vladimir-ch: Can I add these test later? I'd like to merge this first to see some motivating progress.

vladimir-ch:
I really like how the implementation turned out. I think I addressed all your comments. PTAL @btracey

linsolve/gmres.go
vladimir-ch: I added the zeroing to `reuse`.

linsolve/gmres_test.go
btracey: Sure.

btracey:
I haven't had a chance to check the changes yet, FYI.

vladimir-ch:
Oh, sorry, I misunderstood the situation. Do you still want to go through it and add comments here?

btracey:
I think I do, I'm just really busy this week. I hadn't looked at the tests before if you want them checked.

The thing I note at the moment is that "ApplyGivens" is actually `blas.Drotm`, right?

vladimir-ch:
Both `Drot` and `Drotm` operate on two n-vectors. I think it makes little sense to create two slices with one element just to be able to call one of them.

vladimir-ch:
Actually, I ended up trying to use`Drot` and it looks much better then I expected. So I switched to `Drot` in 0c6f383c78ba17376e0c3c7f397e5c008bbc7f63 and removed `applyGivens`. Thanks for bringing it up.

You can take a look at the tests but just the code is enough. I want to return to the tests later, and extend them so that they use not only Matrix Market data.

btracey:
Just want to make sure I'm not holding anything up.

vladimir-ch:
Not at all. I'm slowly preparing a next PR with the driver function but I'm afraid that submitting it will have to wait until January.

btracey:
That's okay. I just realized we left it somewhat ambiguous so wanted to make sure you weren't waiting on me.

PR #9, opened 2018-02-06
linsolve: add driver function Iterative

vladimir-ch:
The main decision point here is how to represent the linear system. I've opted for a struct `System` with a function `MulVec` and a slice `B` but I'm open to suggestions on how an interface representation could look like.

### DIFF
--- a/linsolve/iterative.go
+++ b/linsolve/iterative.go
@@ -31,6 +31,11 @@ type System struct {
 
 // Settings holds various settings for solving a linear system.
 type Settings struct {
+	// InitX holds the initial guess. If it is nil, the zero vector will be
+	// used, otherwise its length must be equal to the dimension of the
+	// system.
+	InitX []float64
+
 	// Tolerance specifies error tolerance for the final approximate
 	// solution produced by the iterative method. Tolerance must be smaller
 	// than one and greater than the machine epsilon. If it is zero,
@@ -118,7 +123,7 @@ type Stats struct {
 //
 // settings provide means for adjusting parameters of the iterative process.
 // Zero values of the fields mean default values.
-func Iterative(sys System, x []float64, method Method, settings Settings) (*Result, error) {
+func Iterative(dst []float64, sys System, method Method, settings Settings) (*Result, error) {
 	stats := Stats{StartTime: time.Now()}
 
 	if sys.MulVec == nil {
@@ -129,20 +134,30 @@ func Iterative(sys System, x []float64, method Method, settings Settings) (*Resu
 		return nil, errors.New("linsolve: dimension not positive")
 	}
 
+	if dst == nil {
+		dst = make([]float64, dim)
+	}
+	if len(dst) != dim {
+		panic("linsolve: mismatched length of dst")
+	}
+
 	ctx := &Context{
+		X:        dst,
 		Residual: make([]float64, dim),
 		Src:      make([]float64, dim),
 		Dst:      make([]float64, dim),
 	}
-	if x != nil {
-		if len(x) != dim {
+	if settings.InitX != nil {
+		if len(settings.InitX) != dim {
 			panic("linsolve: mismatched length of initial guess")
 		}
-		ctx.X = x
+		copy(ctx.X, settings.InitX)
 		computeResidual(ctx.Residual, sys, ctx.X, &stats)
 	} else {
 		// Initial x is the zero vector.
-		ctx.X = make([]float64, dim)
+		for i := range ctx.X {
+			ctx.X[i] = 0
+		}
 		// Residual b-A*x is then equal to b.
 		copy(ctx.Residual, sys.B)
 	}
--- a/linsolve/iterative.go
+++ b/linsolve/iterative.go
@@ -16,9 +16,9 @@ const defaultTolerance = 1e-8
 // without converging to a solution.
 var ErrIterationLimit = errors.New("linsolve: iteration limit reached")
 
-// MulVecer represents a square matrix A by means of a matrix-vector
+// MulVecToer represents a square matrix A by means of a matrix-vector
 // multiplication.
-type MulVecer interface {
+type MulVecToer interface {
 	// MulVecTo computes A*x or A^T*x and stores the result into dst.
 	MulVecTo(dst []float64, trans bool, x []float64)
 }
@@ -120,7 +120,7 @@ type Stats struct {
 // significantly reduce computation time. Thus, while Iterative has supplied
 // defaults, users are strongly encouraged to adjust these defaults for their
 // problem.
-func Iterative(dst []float64, a MulVecer, b []float64, m Method, settings Settings) (*Result, error) {
+func Iterative(dst []float64, a MulVecToer, b []float64, m Method, settings Settings) (*Result, error) {
 	n := len(b)
 	if n == 0 {
 		panic("linsolve: dimension is zero")
@@ -177,7 +177,7 @@ func Iterative(dst []float64, a MulVecer, b []float64, m Method, settings Settin
 	}, err
 }
 
-func iterate(a MulVecer, b []float64, ctx *Context, settings Settings, method Method, stats *Stats) error {
+func iterate(a MulVecToer, b []float64, ctx *Context, settings Settings, method Method, stats *Stats) error {
 	bNorm := floats.Norm(b, 2)
 	if bNorm == 0 {
 		bNorm = 1
@@ -239,7 +239,7 @@ func NoPreconditioner(dst, rhs []float64, trans bool) error {
 	return nil
 }
 
-func computeResidual(dst []float64, a MulVecer, b, x []float64, stats *Stats) {
+func computeResidual(dst []float64, a MulVecToer, b, x []float64, stats *Stats) {
 	stats.MulVec++
 	a.MulVecTo(dst, false, x)
 	floats.AddScaledTo(dst, b, -1, dst)
--- a/linsolve/iterative.go
+++ b/linsolve/iterative.go
@@ -16,16 +16,15 @@ const defaultTolerance = 1e-8
 // without converging to a solution.
 var ErrIterationLimit = errors.New("linsolve: iteration limit reached")
 
-// System describes a linear system in terms of A*x and A^T*x operations, and
-// the right-hand side.
-type System struct {
-	// MulVec computes A*x or A^T*x and stores the result into dst.
-	// MulVec must not be nil.
-	MulVec func(dst, x []float64, trans bool)
+// MulVecer represents a square matrix A of order n by means of a matrix-vector
+// multiplication.
+type MulVecer interface {
+	// Order returns the order n of the matrix A.
+	Order() int
 
-	// B is the right-hand side vector. Its length determines the dimension
-	// of the system.
-	B []float64
+	// MulVec computes A*x or A^T*x and stores the result into dst.
+	// x and dst will have length equal to Order.
+	MulVec(dst, x []float64, trans bool)
 }
 
 // Settings holds settings for solving a linear system.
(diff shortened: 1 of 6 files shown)


linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+
+const DefaultTolerance = 1e-8
+
+// ErrIterationLimit is returned when a maximum number of iterations were done
+// without converging to a solution.
+var ErrIterationLimit = errors.New("linsolve: iteration limit reached")
+
+// System describes a linear system in terms of A*x and A^T*x operations, and
+// the right-hand side.
+type System struct {
```
btracey: I don't see why the `System` struct is needed. Why not just have `Iterative` take in a `MulVecer` and `B` individually?

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+package linsolve
+
+import (
+	"errors"
+	"time"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+const DefaultTolerance = 1e-8
```
btracey: I'm not convinced this should be exported.

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	// MulVec computes A*x or A^T*x and stores the result into dst.
+	// MulVec must not be nil.
+	MulVec func(dst, x []float64, trans bool)
+
+	// B is the right-hand side vector. Its length determines the dimension
+	// of the system.
+	B []float64
+}
+
+// Settings holds various settings for solving a linear system.
```
btracey: Remove `various`?

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+type Settings struct {
+	// Tolerance specifies error tolerance for the final approximate
```
btracey: s/approximate/(approximate)

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	// solution produced by the iterative method. Tolerance must be smaller
```
btracey: s/smaller/less

instead of machine epsilon, maybe `math.SmallestNonzeroFloat64`? It'll be more clear for most readers, and Go requires IEEE 754

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	// than one and greater than the machine epsilon. If it is zero,
+	// DefaultTolerance will be used.
```
btracey: Move the "if zero" case before the "Tolerance must be smaller..."

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	//
+	// If NormA is not zero, the stopping criterion used will be
+	//  |r_i| < Tolerance * (|A|*|x_i| + |b|),
```
btracey: You should define `r_i`

I'm sure this is typical, but why is this the stopping criterion? Seems weird that the tolerance is related to the value of `x`

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// MaxIterations is the limit on the number of iterations. If it is
+	// zero, a default value of twice the dimension of the system will be
+	// used.
+	MaxIterations int
+
+	// PreconSolve describes a preconditioner solve that stores into dst the
+	// solution of the system
+	//  M dst = rhs, or M^T dst = rhs.
+	// If it is nil, no preconditioning will be used (M is the identity).
```
btracey: "where M is the preconditioning matrix"

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	PreconSolve func(dst, rhs []float64, trans bool) error
+}
+
+// defaultSettings fills zero fields of s with default values.
```
btracey: I think this is a bad idea. I know we do it in places in `optimize`, but I don't think we should. I think we should leave the `Settings` struct unmodified by the call. In particular, this makes using the same settings struct for multiple concurrent calls easy.

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+		s.PreconSolve = NoPreconditioner
+	}
+}
+
+// Result holds the result of an iterative solve.
+type Result struct {
+	// X is the approximate solution.
+	X []float64
+
+	// ResidualNorm is the norm of the final residual. It may not be equal
```
btracey: "ResidualNorm is an approximation to the norm ..."

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	// to the actual residual b-A*x.
+	ResidualNorm float64
+
+	// Stats holds statistics about the iterative solve.
+	Stats Stats
+}
+
+// Stats holds statistics about an iterative solve.
+type Stats struct {
+	// Iterations is the number of iterations done by Method.
```
btracey: performed by Method?

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	Iterations int
+
+	// MulVec is the number of MulVec operations commanded by Method.
+	MulVec int
+
+	// PreconSolve is the number of PreconSolve operations commanded by Method.
+	PreconSolve int
+
+	// StartTime is an approximate time when the solve was started.
```
btracey: Is this a useful thing to report?

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	// Runtime is an approximate duration of the solve.
+	Runtime time.Duration
+}
+
+// Iterative solves the system of n linear equations
+//  A*x = b,
+// where the n×n matrix A and the right-hand side b are represented by sys. The
+// dimension n of the system is determined by the length of sys.B.
+//
+// If x is not nil, its length must be equal to n, and it will be used as an
```
btracey: I would rather see a `dst` as the first entry, and have an `init` value that can be nil. Again, I think it's much better policy to not modify inputs (unless they obviously have to be modified).

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+// initial guess. On return it will contain the approximate solution. If it is
+// nil, the zero vector will be used as an initial guess, and the solution will
+// be returned in Result.X.
+//
+// method is an iterative method used for finding an approximate solution of the
+// linear system. It must not be nil.
```
btracey: Is it bad policy to have a default?

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+//
+// settings provide means for adjusting parameters of the iterative process.
+// Zero values of the fields mean default values.
```
btracey: I would just say "see the Settings documentation for more information"

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	stats.Runtime = time.Since(stats.StartTime)
+	return &Result{
+		X:            ctx.X,
+		ResidualNorm: ctx.ResidualNorm,
+		Stats:        stats,
+	}, err
+}
+
+func iterate(sys System, ctx *Context, settings Settings, method Method, stats *Stats) error {
+	bnorm := floats.Norm(sys.B, 2)
```
btracey: bNorm

linsolve/iterative.go
```diff
@@ -0,0 +1,240 @@
...
+	dim := len(ctx.X)
+	method.Init(dim)
+
+	for {
+		op, err := method.Iterate(ctx)
+		if err != nil {
+			return err
+		}
+
+		switch op {
```
btracey: remove all the extra newlines.

linsolve/iterative.go
vladimir-ch: I had to reduce the number of parameters and to avoid mistakes from confused slices b and x. I have now removed `System` and added `dst`, `Settings.InitX`, and MulVecer interface.

linsolve/iterative.go
vladimir-ch: Unexported now.

linsolve/iterative.go

linsolve/iterative.go

linsolve/iterative.go
vladimir-ch: Done. I wonder whether we should do the same in `optimize`?

linsolve/iterative.go
vladimir-ch: We could have GMRES as a default but my objective with this was to force the uder to think about the problem, its properties and about the solver being used.

linsolve/iterative.go

linsolve/iterative.go
vladimir-ch: Done. Also rNorm and xNorm.

linsolve/iterative.go

vladimir-ch:
I like MulVec but I'm not so happy about the name conflict with mat.VecDense.MulVec.

I've addressed the comments, PTAL.

linsolve/iterative.go

linsolve/iterative.go

The Templates book provides some motivation and theory behind this (and other) stopping criterion.

linsolve/iterative.go

linsolve/iterative.go
vladimir-ch: `Iterative` itself takes `Settings` by value which requires to pass them explicitly always but it allows me to modify them freely.

linsolve/iterative.go

linsolve/iterative.go

linsolve/iterative.go
vladimir-ch: Not really. Removed and `Runtime` as well.

linsolve/iterative.go
```diff
@@ -0,0 +1,237 @@
...
+
+// MulVecer represents a square matrix A of order n by means of a matrix-vector
+// multiplication.
+type MulVecer interface {
+	// Order returns the order n of the matrix A.
+	Order() int
+
+	// MulVec computes A*x or A^T*x and stores the result into dst.
+	// x and dst will have length equal to Order.
+	MulVec(dst []float64, trans bool, x []float64)
```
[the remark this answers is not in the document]
vladimir-ch: Simple and effective distinction. Done.

linsolve/iterative.go
```diff
@@ -0,0 +1,237 @@
...
+	// PreconSolve is the number of PreconSolve operations commanded by Method.
+	PreconSolve int
+}
+
+// Iterative finds an approximate solution of the system of n linear equations
+//  A*x = b,
+// where A is a square matrix of order n and b is the right-hand side vector,
+// using an iterative method m.
+//
+// If dst is not nil, its length must be equal to n and the result will be
```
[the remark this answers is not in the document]

linsolve/iterative.go
```diff
@@ -0,0 +1,237 @@
...
+	Iterations int
+
+	// MulVec is the number of MulVec operations commanded by Method.
+	MulVec int
+
+	// PreconSolve is the number of PreconSolve operations commanded by Method.
+	PreconSolve int
+}
+
+// Iterative finds an approximate solution of the system of n linear equations
```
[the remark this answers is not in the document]
vladimir-ch: In mathematical context I think I've never seen "solution for the system/equation/problem", it's always "of". I've checked several books from different authors and "solution of the/a system" is used always. I would use "for" in a context such as "obtain solution for a coarse grid".

linsolve/iterative.go
```diff
@@ -0,0 +1,237 @@
...
+// using an iterative method m.
+//
+// If dst is not nil, its length must be equal to n and the result will be
+// stored into dst, otherwise a new slice will be allocated and returned in
+// Result.
+//
+// settings provide means for adjusting parameters of the iterative process.
+// See the Settings documentation for more information.
+func Iterative(dst []float64, a MulVecer, b []float64, m Method, settings Settings) (*Result, error) {
+	n := a.Order()
```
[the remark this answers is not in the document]
vladimir-ch: I agree. Removed.

linsolve/iterative.go
```diff
@@ -0,0 +1,237 @@
...
+// Iterative finds an approximate solution of the system of n linear equations
+//  A*x = b,
+// where A is a square matrix of order n and b is the right-hand side vector,
+// using an iterative method m.
+//
+// If dst is not nil, its length must be equal to n and the result will be
+// stored into dst, otherwise a new slice will be allocated and returned in
+// Result.
+//
+// settings provide means for adjusting parameters of the iterative process.
```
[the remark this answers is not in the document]
vladimir-ch: Method can now be nil, and GMRES will be used as a default. However, this also implies a default value for GMRES.Restart. Now it is `n` but that was chosen assuming that the user has access to the Restart value and will change it if necessary. In this new situation with GMRES as a convenient default I think `n` is a bad and dangerously large value for Restart. Setting it to e.g. 10 does not converge for 4 out of 5 unsymmetric test cases. On one hand this can happen and the test matrices will not be particularly nice but if I removed them there would be only one unsymmetric test case left (also it could also indicate a bug in the restarting logic/indexing).

I will leave it as it is now (Restart = n) but will come back to it later. This is an experimental repository after all.

And thanks for the paragraph, nicely written.

linsolve/iterative.go
```diff
@@ -0,0 +1,237 @@
...
+
+	if dst == nil {
+		dst = make([]float64, n)
+	}
+	if len(dst) != n {
+		panic("linsolve: mismatched length of dst")
+	}
+
+	var stats Stats
+	ctx := &Context{
```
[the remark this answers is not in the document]
vladimir-ch: This is an interesting idea. At the moment I will not do it, I need to think about how `Settings.{InitX,Dst,Context}` and `Result` would work together. Instead of user providing Context they could give a large Lapack-style work slice.

linsolve/iterative.go
```diff
@@ -0,0 +1,237 @@
...
+	// If NormA is zero (not available), the stopping criterion will be
+	//  |r_i| < Tolerance * |b|.
+	Tolerance float64
+
+	// NormA is an estimate of a norm of the matrix A. For example,
+	// an approximation of the absolute value of the largest entry of A can be
+	// used. Zero value means that the norm is not known.
+	NormA float64
+
+	// MaxIterations is the limit on the number of iterations. If it is
```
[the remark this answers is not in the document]
vladimir-ch: Yea, it's difficult, I don't think there is a typical default. It is even more complicated by the fact that what an iteration is depends entirely on Method. Moreover, theoretically any convergence profile is possible for e.g. GMRES so initially it may reduce the residual norm slowly or not at all and converge faster only later, or even only in the last (n-th) iteration. In practice (floating point) the convergence may be delayed even more.

Either way, I view `MaxIterations` not as a kind of stopping criterion but rather as a device to avoid infinite iteration loop when unrealistically low tolerance is used. So yes, in practical situations I would of course expect convergence much earlier.

vladimir-ch:
I think that you can take a look again.

linsolve/iterative.go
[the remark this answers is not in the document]
btracey: Okay, it's fine as is then.

vladimir-ch:
Thanks for the review.

linsolve/iterative.go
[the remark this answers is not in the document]
vladimir-ch: Regarding the test failures when Restart != n, I get the same failures (no convergence, no progress) also with Octave, so I think there is no bug, it's just a loss of convergence which can sometimes occur when using restarting. But at least it suggests that I should add detection of no progress in the iterates (which can be often detected much earlier than iteration limit is exhausted), so that will be the next PR. I'll also take a look on the tests but as with optimize, testing these iterative algorithms is hard.

PR #10, opened 2018-02-18
Sparse level1

rlouf:
I have implemented the level1 Sparse BLAS vector operations as specified [here](http://www.netlib.org/blas/blast-forum/chapter3.pdf). But:

- I saw in the [gonum repository](https://github.com/gonum/gonum/blob/master/blas/gonum/level1double.go) that some rotation operations are also defined for dense matrices.
- I implemented `Swap` a `x, y = y, x` and wondered if I should have done this "by hand"?

The tests pass, but I need to add more test cases.

I would love to have some feedback on what I've done so far. What doesn't look good, what's wrong and what should be added next. 

Thanks!


sparse/level1.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2018 The Gonum Authors. All rights reserved.
```
kortschak: `// Copyright ©2018 The Gonum Authors. All rights reserved.`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+// Use of this source code is governed by BSD-style
+// license that can be found in the LICENSE file.
+//
+// Functions in this file implement the Level 1 BLAS specifications
```
kortschak:
```
// Package sparse implements Level 1 BLAS for sparse vectors.
```

Though this should really be in a doc.go file.

It would be good to provide a link to the spec that you are working from too.

(Although I guess you could leave out "Level 1" since we are aiming for BLAS for sparse vectors).

Also, please ensure there is no blank line between the package-level comment and the package declaration - godoc uses this to generate on-line documentation.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+// for sparse vectors
+//
+// Dot: dot product between sparse vectors
```
kortschak: These can be distributed across the relevant functions; godoc will generate an index for us.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+package blas64
+
+import (
+	"math"
+)
+
+// SparseVector represents a sparse vector.
+type SparseVector struct {
+	Size int // Number of elements in the (non sparse) vector.
+	Ind  []int
```
kortschak: Please describe the organisation/invariants of `Ind` and `Data` here.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+	Data []float64
+}
+
+type DenseVector struct {
```
kortschak: Can you use a blas64.Vector here instead please (unless there is something in the BLAS sparse spec that prohibits it).

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+	Data []float64
+}
+
+// Level 1 --------------------------------------------------------------------
```
kortschak: We eschew this kind of comment; it is preferable to split the files rather than add decoration.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+
+//  Dot computes the dot product between two sparse vectors
+//  Dot(x, y) = \sum_i x[i]*y[i]
+//  Will panic if vectors are not the same size
+func Dot(x, y SparseVector) float64 {
+	if x.Size != y.Size {
+		panic("X and Y must have the same length.")
```
kortschak: `"sparse: mismatched sparse vector sizes"`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+	return dot
+}
+
+// Nrm2 computes the L2 norm of the vector x:
+// Nrm2(x) = \sum_i x[i]*x[i]
+func Nrm2(x SparseVector) float64 {
+	var nrm2 float64
+	for _, value := range x.Data {
```
kortschak: `s/value/v/g`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+		nrm2 += value * value
+	}
+	return math.Sqrt(nrm2)
+}
+
+// Asum computes the sum of the absolute values of the elemnts of x:
+// Asum(x) = \sum_i |x[i]|
+func Asum(x SparseVector) float64 {
+	var asum float64
+	for _, value := range x.Data {
```
kortschak: `s/value/v/g`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+
+type DenseVector struct {
+	Data []float64
+}
+
+// Level 1 --------------------------------------------------------------------
+
+//  Dot computes the dot product between two sparse vectors
+//  Dot(x, y) = \sum_i x[i]*y[i]
+//  Will panic if vectors are not the same size
```
kortschak: `// Dot will panic`...

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return asum
+}
+
+// Idmax returns the index of an element of x with the largest aboslute value
+// If there are multiple such indices the maximum is returned.
+// Idmax returns -1 if n == 0
+func Idmax(x SparseVector) int {
+	if x.Size == 0 {
+		return 0
```
kortschak: Is this what the spec asks for? A non-negative flag value for a zero-size vector seems foolhardy.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+// Asum(x) = \sum_i |x[i]|
+func Asum(x SparseVector) float64 {
+	var asum float64
+	for _, value := range x.Data {
+		asum += math.Abs(value)
+	}
+	return asum
+}
+
+// Idmax returns the index of an element of x with the largest aboslute value
```
kortschak: `s/aboslute/absolute/`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+// If there are multiple such indices the maximum is returned.
+// Idmax returns -1 if n == 0
```
kortschak: What is `n` here?

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+func Asum(x SparseVector) float64 {
+	var asum float64
+	for _, value := range x.Data {
+		asum += math.Abs(value)
+	}
+	return asum
+}
+
+// Idmax returns the index of an element of x with the largest aboslute value
+// If there are multiple such indices the maximum is returned.
```
kortschak: This is not the code's behaviour, do you mean "the first"?

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+			max = x.Data[i]
+		}
+	}
+	return imax
+}
+
+// Swap exchanges the elements of the two vectors x and y
+// x[i], y[i] = y[i], x[i] for all i
+// Will panic if the vectors are not of the same length
+func Swap(x, y *SparseVector) {
```
kortschak: Is this how the reference implements this! I don't like this at all. The elements should be moved.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+			imax = ind
+			max = x.Data[i]
+		}
+	}
+	return imax
+}
+
+// Swap exchanges the elements of the two vectors x and y
+// x[i], y[i] = y[i], x[i] for all i
+// Will panic if the vectors are not of the same length
```
kortschak: `// Swap will panic`...

Also, final full stop.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+func Swap(x, y *SparseVector) {
+	if x.Size != y.Size {
+		panic("input vectors are not of the same size")
```
kortschak: `"sparse: mismatched sparse vector sizes"`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	x, y = y, x
+}
+
+// Scal scales the vector x by \alpha:
+// x[i] *= \alpha for all i
+func Scal(alpha float64, x SparseVector) SparseVector {
+	if alpha == 0 {
+		return SparseVector{x.Size, []int{}, []float64{}}
```
kortschak: There is no need to return a value; BLAS works in place.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+
+	for i, _ := range x.Data {
+		x.Data[i] *= alpha
+	}
+	return x
+}
+
+// Copy copies the elements of x into the elements of y
+// y[i] = x[i] for all i
+// Will panic if x and y are not of same length
```
kortschak: `// Copy will panic`... and full stop.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+func Copy(x, y *SparseVector) {
+	if x.Size != y.Size {
+		panic("X and Y must have the same length.")
```
kortschak: `"sparse: mismatched sparse vector sizes"`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	y.Ind = y.Ind[:0]
```
kortschak: Can you not do 
```
y.Ind = append(y.Ind[:0], x.Ind...)
y.Data = append(y.Data[:0], x.Data...)
```

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+	y.Data = y.Data[:0]
+	for i, ind := range x.Ind {
+		y.Ind = append(y.Ind, ind)
+		y.Data = append(y.Data, x.Data[i])
+	}
+}
+
+// Axpy add alpha time x to y
+// y[i] += alpha * x[i] for all i
+// Current version does not keep an ordered vec
```
kortschak: Does the spec have this order of parameters? It's unusual.

Also, you need to clarify the implication of the index ordering here.

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		ity += 1
+	}
+}
+
+// Gather takes a DenseVector as an input and
+// transforms it in a sparse vector
+func Gather(x DenseVector) SparseVector {
+	var y SparseVector
+	for i, datum := range x.Data {
```
kortschak: `s/datum/v/g`

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+		if i == j {
+			y.Data[ity] += alpha * x.Data[itx]
+		}
+		ity += 1
+	}
+}
+
+// Gather takes a DenseVector as an input and
+// transforms it in a sparse vector
+func Gather(x DenseVector) SparseVector {
```
kortschak: Does the spec state what form this function should take? If not, I'd prefer this were `Gather(dst *SparseVector, x DenseVector)` (for whatever type the dense vector is).

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+		if datum != 0 {
+			y.Ind = append(y.Ind, i)
+			y.Data = append(x.Data, datum)
+		}
+	}
+	return y
+}
+
+// Scatter builds a DenseVector from a SparseVector
+func Scatter(y SparseVector) DenseVector {
```
kortschak: Similar here.

If a blas64.Vector is used, a non-zero `Inc` should be adhered to.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
+package blas64
```
kortschak: Copyright header here too please.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+import (
+	"testing"
+)
+
+// Single vector operations ---------------------------------------------------
+type OneVectorCase struct {
```
kortschak: This does not need to be exported, for any of the types and fields here this is true.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
+package blas64
+
+import (
+	"testing"
+)
+
+// Single vector operations ---------------------------------------------------
```
kortschak: We don't need this.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+	Panic     bool
+	Asum      float64
+	Nrm2      float64
+	Idmax     int
+	ScalCases []ScalCase
+}
+
+type ScalCase struct {
+	Alpha  float64
+	Answer SparseVector // Values of the sparse vector
```
kortschak: `s/Answer/want/g`

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+	Name   string
+}
+
+var OneVectorCases = []OneVectorCase{
+	{
+		Name:   "Dense All Positive",
+		Vector: SparseVector{5, []int{0, 1, 2, 3, 4}, []float64{6, 5, 4, 2, 6}},
+		Panic:  false,
+		Asum:   23,
+		Nrm2:   10.81665382639196787935766380241148783875388972153573863813135,
```
kortschak: This is greater precision than will be possible for `float64`, so it's not really necessary.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+				Answer: SparseVector{5, []int{0, 1, 2, 3, 4}, []float64{-12, -10, -8, -4, -12}},
+			},
+		},
+	},
+	{
+		Name:   "Sparse All Positive",
+		Vector: SparseVector{15, []int{1, 3, 4, 10, 14}, []float64{6, 5, 4, 2, 6}},
+		Panic:  false,
+		Asum:   23,
+		Nrm2:   10.81665382639196787935766380241148783875388972153573863813135,
```
kortschak: Same here.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+				Answer: SparseVector{15, []int{1, 3, 4, 10, 14}, []float64{-12, -10, -8, -4, -12}},
+			},
+		},
+	},
+	{
+		Name:   "Dense All negative",
+		Vector: SparseVector{5, []int{0, 1, 2, 3, 4}, []float64{-6, -5, -4, -2, -6}},
+		Panic:  false,
+		Asum:   23,
+		Nrm2:   10.81665382639196787935766380241148783875388972153573863813135,
```
kortschak: And here.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+			},
+			{
+				Alpha:  -2,
+				Answer: SparseVector{0, []int{}, []float64{}},
+			},
+		},
+	},
+}
+
+// Two vectors operations -----------------------------------------------------
```
kortschak: Remove this please.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+			{
+				Name:  "Size, Ind and Data Matching",
+				Panic: false,
+				Input: SparseVector{5, []int{0, 1, 1, 1, 1}, []float64{0, 3, 4, 5, 1}},
+			},
+		},
+	},
+}
+
+// Mixed vectors operations ---------------------------------------------------
```
kortschak: And this.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+var MixedVectorsCases = []MixedVectorsCase{
+	{
+		Name:   "Dense to dense",
+		Sparse: SparseVector{5, []int{0, 1, 2, 3, 4}, []float64{0, 4, 6, 3, 0}},
+		Dense:  DenseVector{[]float64{0, 4, 6, 3, 0}},
+		Panic:  false,
+	},
+}
+
+// Test single vector operations ----------------------------------------------
```
kortschak: And here.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+					if val != scal.Answer.Data[i] {
+						t.Errorf("ascal: mismatch of data in %v: expected %v, found %v", c.Name, scal.Answer.Data[i], val)
+					}
+				}
+			}
+		}
+	}
+}
+
+// Test 2 sparse vectors cases ------------------------------------------------
```
kortschak: And here.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+			// Test
+			Axpy(&c.Vector1, &output, cal.Alpha)
+			if !isEqualSparse(output, cal.Answer) {
+				t.Errorf("axpy: %s: %s: output does not match the expected vector", c.Name, cal.Name)
+			}
+		}
+	}
+}
+
+// Test 1 sparse and 1 dense vectors cases ------------------------------------
```
kortschak: And here.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+func testFromDense(t *testing.T) {
```
kortschak: These tests won't be run since the test func is not `Test`... Please make the name match the function being tested, also below.

sparse/level1_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+func testScatter(t *testing.T) {
+	for _, c := range MixedVectorsCases {
+		output := Scatter(c.Sparse)
+		if !isEqualDense(output, c.Dense) {
+		}
+	}
+}
+
+// Helper functions -----------------------------------------------------------
```
kortschak: Remove this please.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Took me a couple of minute to figure these were `vi` commands :) Isn't it always better to be explicit when you can? It doesn't impact readability so much.

sparse/level1.go
rlouf: Same comment.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Solved.

rlouf:
Thanks for thoroughly checking my code! I'll follow the specifications more closely --- what baffled me was that the `Copy` and `Swap` functions where neither taking pointers as an input, nor return anything. Maybe there is something I'm missing?

I'm going to make the required changes and push them here.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
rlouf: Solved.

sparse/level1.go
kortschak: Conventionally, we use short labels. This one live for two lines, so it is easy to see what it is.

Please take a look at https://talks.golang.org/2014/names.slide, https://research.swtch.com/names and the variable names section of https://www.lysator.liu.se/c/pikestyle.html.

kortschak:
>Maybe there is something I'm missing?

Can you point me to the spec you're reading from?

sparse/level1.go
```diff
@@ -0,0 +1,190 @@
...
+	Data []float64
+}
+
+type DenseVector struct {
+	Data []float64
+}
+
+// Level 1 --------------------------------------------------------------------
+
+//  Dot computes the dot product between two sparse vectors
```
vladimir-ch: I think we should first clarify what will go into our sparse BLAS. Sparse BLAS reference linked to above covers at "level 1" only the dot product, axpy, gather, and scatter. Also it covers no sparse-sparse operations, so for example in case of `Dot` it would mean `Dot(x SparseVector, y blas64.Vector) float64`.

rlouf: Let's agree that we begin with a strict implementation of the BLAS reference, then. Having said that, I very often need sparse-sparse operations and would like to see them implemented at some point. We can always discuss that once we have all the basics covered.

sparse/level1_test.go
rlouf: Solved

sparse/level1_test.go
rlouf: Solved

sparse/level1_test.go
rlouf: Solved

sparse/level1_test.go
rlouf: Solved.

kortschak:
As pointed out by @vladimir-ch, the Sparse BLAS reference does not include the full dense BLAS, so please work from the Sparse BLAS spec, not our BLAS implementation.

sparse/level1.go
kortschak: Sparse/sparse is done with dense working space. We can handle that later, when the spec-defined functions are provided here.

kortschak:
I would recommend that you read both the SparseBLAS spec and the CSparse book to get a good idea of what happens and why (see [here](https://github.com/gonum/matrix/issues/301#issuecomment-164599167) for links).

rlouf:
The Csparse book is mostly about matrices (the difficult stuff), but I’ll stick to the specs from now on. Btw, before I do, is it ok to look at the Csparse code—license-wise?

kortschak:
The CSparse book is a teaching book, so what you learn from reading it is not a licensed thing (any legal view that this is not true would put all the world's post enlightenment learning/intellectual development at risk). The C code that goes along with the book is GLPL licensed, so we can't include it in the Gonum, but the code there is very specifically C, and I doubt anyone would confuse a Go sparse BLAS implementation with how CSparse is written. If in doubt, only read the book (note that the code is really quite subtle, and a translation would likely be more difficult than an _ab_initio_ implementation).

Having said all this, reading the book would be useful since it does cover all the subtleties of sparse representation (including of vectors and the gather/scatter operations).

rlouf:
Following all the above discussions, I decided to strictly follow the Sparse BLAS specification (as detailed here http://www.netlib.org/blas/blast-forum/chapter3.pdf). The result can be found in level1double.go. If this looks ok to you I will proceed with writing tests and moving on to the matrices.

sparse/level1double.go
```diff
@@ -0,0 +1,60 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by BSD-style
+// license that can be found in the LICENSE file.
+package spblas64
```
kortschak: At this point, I think the package becomes sblas without the float size determinant. We can code-generate the single precision version from this.

sparse/level1double.go
```diff
@@ -0,0 +1,60 @@
...
+
+// SDot computes the dot product between a sparse vector x and a dense vector y
+// SDot(x, y) = \sum_i x[i]*y[index[i]*incY]
+func SDot(nz int, x []float64, index []int, y []float64, incY int) float64 {
```
kortschak: Can you use the Sparse BLAS names, e.g. this would be `Dusdot`.

sparse/level1double.go
```diff
@@ -0,0 +1,60 @@
...
+	var dot float64
+	for i = 0; i < nz; i++ {
+		dot += x[i] * y[index[i]*incY]
```
kortschak: incY can be negative, which reverses the direction of the operation. That would panic here.

Have a look at how this is [done in blas](https://github.com/gonum/gonum/blob/master/blas/gonum/level1double_ddot.go#L13-L49). Though it's not so important for dot that the direction match, but it should not panic.

sparse/level1double.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	return dot
+}
+
+// SAxpy adds alpha time the sparse vector x to the dense vector y
+// y[index[i]*incY] += alpha * x[i] for all i
+func SAxpy(nz int, alpha float64, x []float64, index []int, y *[]float64, incY int) {
```
kortschak: Please make a similar name change.

sparse/level1double.go
```diff
@@ -0,0 +1,60 @@
...
+	for i = 0; i < nz; i++ {
+		y[index[i]*incY] += alpha * x[i]
```
kortschak: incY here can be negative.

sparse/level1double.go
```diff
@@ -0,0 +1,60 @@
+// Copyright ©2018 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by BSD-style
+// license that can be found in the LICENSE file.
+package spblas64
+
+// SDot computes the dot product between a sparse vector x and a dense vector y
+// SDot(x, y) = \sum_i x[i]*y[index[i]*incY]
```
kortschak: I think that probably the high level interpretation is what should be here, rather than the by-indexed-element explanation, so r = x^T y rather than what is here.

sparse/level1double.go
rlouf: Should I separate the part that checks the input and the one that does the actual computation as is already the case in blas? Have the equivalent of `blas/gonum` and `internal`? 
Things quickly get messy when you add all those guards.

sparse/level1double.go
kortschak: Yeah, that notation in blas is pseudo indexing; the same notation would be used here in that it's referring to the conceptual vectors rather than our implementation of them.

sparse/level1double.go
kortschak: The internal implementations are there because they can have an asm implementation. If/when we have an asm implementation of this, we can move the kernels out to that, but at this stage, just leave them in the body proper. Don't worry about the apparent "mess" of checks, the explicitness of checking is important for us.

rlouf: Ok! Naive questions: why do we keep the possibilty to have negative `incY` in Go? Is it a story of compatibility with old code in C for people who are switching?

PR #11, opened 2018-04-12
linsolve: rework Settings

vladimir-ch:
This PR makes changes suggested by @btracey : 
* move `dst` parameter into `Settings`
* add `Context` to `Settings` so that it can be reused to reduce memory allocations.

There are also some other related changes like passing `Settings` to `Iterative` by pointer.

I'm quite happy with the changes and how they simplify the API of `Iterative`.

I've also tried adding detection when a method is not making any progress but in the end decided against it. The check would have to have a tolerance, but even if it was the machine epsilon, the convergence pattern is in general unpredictable and this could lead to false positives. I left the commits there for reference.

### DIFF
--- a/linsolve/iterative.go
+++ b/linsolve/iterative.go
@@ -70,10 +70,14 @@ type Settings struct {
 	// where M is the preconditioning matrix. If PreconSolve is nil, no
 	// preconditioning will be used (M is the identity).
 	PreconSolve func(dst, rhs []float64, trans bool) error
+
+	// Work context can be provided to reduce memory allocation when solving
+	// multiple linear systems.
+	Work *Context
 }
 
 // DefaultSettings returns default settings for solving a general linear system
-// of dimension dim. It will not allocate fields like InitX or Dst.
+// of dimension dim. It will not allocate fields like InitX, Dst, or Work.
 func DefaultSettings(dim int) *Settings {
 	var s Settings
 	defaultSettings(&s, dim)
@@ -140,7 +144,8 @@ type Stats struct {
 //
 // settings provide means for adjusting parameters of the iterative process. The
 // nil value of settings is equivalent to using DefaultSettings. See the Settings
-// documentation for more information.
+// documentation for more information. Iterative will not modify the fields of
+// Settings.
 //
 // Note that the default choices of Method and Settings were chosen to provide
 // accuracy and robustness, rather than speed. There are many algorithms for
@@ -168,7 +173,7 @@ func Iterative(a MulVecToer, b []float64, m Method, settings *Settings) (*Result
 	checkSettings(&s, n)
 
 	var stats Stats
-	ctx := newContext(n)
+	ctx := reuseContext(s.Work, n)
 	if s.InitX != nil {
 		copy(ctx.X, s.InitX)
 		computeResidual(ctx.Residual, a, b, ctx.X, &stats)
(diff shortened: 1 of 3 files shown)


linsolve/iterative.go
```diff
@@ -58,6 +64,18 @@ type Settings struct {
 	// where M is the preconditioning matrix. If PreconSolve is nil, no
 	// preconditioning will be used (M is the identity).
 	PreconSolve func(dst, rhs []float64, trans bool) error
+
+	// Work context can be provided to reduce memory allocation when solving
+	// multiple linear systems.
+	Work *Context
+}
+
+// DefaultSettings returns default settings for solving a general linear system
```
btracey: I don't think there's any need for this. The default value is the zero value

linsolve/iterative.go
```diff
@@ -73,6 +91,21 @@ func defaultSettings(s *Settings, dim int) {
 	}
 }
 
+func checkSettings(s *Settings, dim int) {
```
btracey: Add in context check.

linsolve/iterative.go
```diff
@@ -120,58 +150,45 @@ type Stats struct {
...
-		Residual: make([]float64, n),
-		Src:      make([]float64, n),
-		Dst:      make([]float64, n),
-	}
-	if settings.InitX != nil {
-		if len(settings.InitX) != n {
-			panic("linsolve: mismatched length of initial guess")
-		}
-		copy(ctx.X, settings.InitX)
+	ctx := reuseContext(s.Work, n)
```
btracey: See below.

linsolve/linsolve.go
```diff
@@ -65,6 +65,30 @@ type Context struct {
...
+	}
+	return &Context{
+		X:        make([]float64, n),
+		Residual: make([]float64, n),
+		Src:      make([]float64, n),
+		Dst:      make([]float64, n),
+	}
+}
+
+func reuseContext(ctx *Context, n int) *Context {
```
btracey: I think that this should be exported, and instead we should force that `Context` is the right size when passed in by the user.

linsolve/iterative.go
vladimir-ch: Right. Done.

linsolve/iterative.go

linsolve/linsolve.go
vladimir-ch: The reasoning was to not panic if it's only optional and not directly related to system solving. But I'm happy to change, it make the rules simpler. Done.

vladimir-ch:
PTAL

linsolve/iterative.go
```diff
@@ -71,6 +85,43 @@ func defaultSettings(s *Settings, dim int) {
...
+	if s.Dst != nil && len(s.Dst) != dim {
+		panic("linsolve: mismatched destination length")
+	}
+	if s.Tolerance <= 0 || 1 <= s.Tolerance {
+		panic("linsolve: invalid tolerance")
+	}
+	if s.MaxIterations <= 0 {
+		panic("linsolve: negative iteration limit")
+	}
+	if s.Work != nil && !checkContext(s.Work, dim) {
```
btracey: I don't think s.Work can be nil because of the call to default settings first. Seems like this should just checkContext

vladimir-ch:
> For instance, I don't quite see why the removed zeroing of ctx.X doesn't need to be there any more.

Good catch. In one of the intermediate commits it was not necessary so I removed it. I will add a test for context containing garbage.

linsolve/iterative.go
vladimir-ch: Done, and also for `Dst`.

PR #12, opened 2018-06-26
linsolve: rework testing

vladimir-ch:
The tests for `linsolve` rely mostly on matrices from Matrix Market (MM) with some limited additional generated matrices. While MM matrices have a variety of properties and origins, on the downside they are often very ill-conditioned. This makes the tests difficult to pass, requires fiddling with tolerances and iteration limits, and often the convergence is very slow and the tests take long time to finish. Since I want the tests to be fast and I don't want to test the methods per se, just that their implementation is correct, I prefer a set of easy generated test matrices.

This PR therefore removes the MM tests and data, and replaces them with generated matrices. Some matrices are random with given properties (`newRandomSPD`, `newRandomDiagonal`), some are random with prescribed eigenvalue distribution (`newGreenbaum41`, `newGreenbaum54`), but most come from a discretization of a PDE (`newGreenbaum73`, `newPoisson*D`, `newPDE*`).

There is also included a change how convergence check is commanded by Methods and what MajorIteration means. Further description is in the corresponding commit.

Post hoc I split the change into several commits for easier review but the main changes are still in only one commit where the diff is useless. `linsolve_test.go` and `iterative_test.go` are best reviewed as they are.

Please take a look.


linsolve/iterative_test.go
```diff
@@ -6,71 +6,146 @@ package linsolve
...
+		Dst:           dst,
+		Tolerance:     tc.tol,
+		MaxIterations: 3 * n,
+		PreconSolve:   tc.PreconSolve,
+		Work:          work,
+	}
+}
+
+func testMethodWithSettings(t *testing.T, m Method, s *Settings, tc testCase) {
+	wantTol := 1e-11
```
btracey: I don't understand these lines. Why is the tolerance lower if settings isn't specified? If it's about the different cases being easier to converge, maybe this should be an input?

linsolve/linsolve_test.go
```diff
@@ -5,104 +5,725 @@
...
+//  - Δu = f
+// on the unit square [0,1]×[0,1].
+func newPoisson2D(nx, ny int, rnd *rand.Rand) testCase {
+	tc := newPDE(nx, ny, negOne, negOne, zero, zero, zero, random(rnd))
+	tc.name = fmt.Sprintf("Poisson 2D nx=%v,ny=%v", nx, ny)
+	tc.tol = 1e-14
+	return tc
+}
+
+// newRandomUnsym returns a test case with a general unsymmetric matrix
```
btracey: Function name does not match the comment name.

linsolve/iterative_test.go
vladimir-ch: I have added a comment that explains this. The individual test cases use very low convergence tolerance, one which is still attainable for the test case (and for the two methods I have so far). If default Settings are used, the tolerance is higher, so the tolerance for the accuracy of the computed solution must be also increased. In other words, it is not about the different test cases being easier to converge.

linsolve/linsolve_test.go
vladimir-ch: Thanks, fixed now.

vladimir-ch:
> I didn't carefully check the matrix examples, but I can you'd like.

If you're busy, you can just skim through them to see whether they are clear enough.

vladimir-ch:
I think that's ok for now, I will check them as I keep working on the package.

Thanks for the review.

PR #19, opened 2019-01-11
Dframe proposal

sbinet:


dframe/README.md
```diff
@@ -0,0 +1,177 @@
...
+func (df *Frame) NumCols() int { ... }
+
+// Column returns the i-th column of this Frame.
+func (df *Frame) Column(i int) *array.Column { ... }
+
+// ColumnNames returns the list of column names of this Frame.
+func (df *Frame) ColumnNames() []string { ... }
+```
+
+It is expected to build `dframe.Frame` on top of `arrow/array.Interface` and/or `arrow/tensor.Interface` to re-use the SIMD optimized operations and zero-copy optimization that are implemented within these packages.
```
Soypete: why are you choosing to have the frames be dependent on the arrow array and tensor interfaces?

dframe/README.md
```diff
@@ -0,0 +1,177 @@
...
+func (df *Frame) NumRows() int { ... }
+
+// NumCols returns the number of columns of this Frame.
+func (df *Frame) NumCols() int { ... }
+
+// Column returns the i-th column of this Frame.
+func (df *Frame) Column(i int) *array.Column { ... }
+
+// ColumnNames returns the list of column names of this Frame.
+func (df *Frame) ColumnNames() []string { ... }
```
kortschak: This is discordant with the `Column` method.

dframe/README.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+It is expected to build `dframe.Frame` on top of `arrow/array.Interface` and/or `arrow/tensor.Interface` to re-use the SIMD optimized operations and zero-copy optimization that are implemented within these packages.
+Using Arrow should also allow seamless interoperability with other data wrangling systems, possibly written in other languages than Go.
+
+`tobgu/qframe` presents a `QFrame` type that is essentially immutable.
+Operations on a `QFrame`, such as copying columns, dropping columns, sorting them or applying some kind of operation on columns, return a new `QFrame`, leaving the original untouched.
+
+Arrow uses a ref-counting mechanism for all the types that involve memory allocation (mainly to address workloads involving memory allocated on a GPGPU, by a SQL database or a mmap-file.)
+This ref-counting mechanism is presented to the user as a pair of methods `Retain/Release` that resp. increment and decrement that reference count.
```
kortschak: s!`Retain/Release`!`Retain`/`Release`!
d/resp. /

dframe/README.md
```diff
@@ -0,0 +1,177 @@
...
+At first, it would seem this mechanism would prevent to expose an API with "chained methods", as the intermediate `Frame` would be "leaked":
```
kortschak: This feels like it's leading towards a "However, ...", but that doesn't arrive until lower down after the code. I think there needs to be clarification earlier.

dframe/dfmat/dfmat.go
```diff
@@ -0,0 +1,142 @@
...
+func FromMatrix(m mat.Matrix, opts ...Option) *dframe.Frame {
+	var (
+		mem = memory.NewGoAllocator()
+		bld = array.NewFloat64Builder(mem)
+
+		r, c   = m.Dims()
+		arrs   = make([]array.Interface, c)
+		fields = make([]arrow.Field, c)
+	)
+	defer bld.Release()
```
kortschak: This feels oddly distant from the `bld` decl.

dframe/dframe.go
```diff
@@ -9,5 +9,620 @@
...
+	md := tx.df.Schema().Metadata() // FIXME(sbinet): also remove metadata of removed cols.
+	sc := arrow.NewSchema(fs, &md)
+
+	tx.df.cols = cs
+	tx.df.schema = sc
+	return tx
+}
+
+var (
+	_ array.Table = (*Frame)(nil)
```
kortschak: Can you move this up to near the type decl for `Frame`. This is a contract documentation, so it should be visible early. If there is only one, just `var array.Table = (*Frame)(nil)`.

dframe/dframe_example_test.go
```diff
@@ -0,0 +1,343 @@
...
+	"log"
+
+	"github.com/apache/arrow/go/arrow"
+	"github.com/apache/arrow/go/arrow/array"
+	"github.com/apache/arrow/go/arrow/memory"
+	"gonum.org/v1/exp/dframe"
+)
+
+func ExampleFrame_fromTable() {
+
```
kortschak: Delete blank line.

dframe/dframe_example_test.go
```diff
@@ -0,0 +1,343 @@
...
+	// rec[1]["f2-f64"]: [6 7 8 9 10]
+	// rec[1]["fx-f64"]: [6 7 8 9 10]
+	// rec[2]["f2-f64"]: [11 12 13 14 15]
+	// rec[2]["fx-f64"]: [11 12 13 14 15]
+	// rec[3]["f2-f64"]: [16 17 18 19 20]
+	// rec[3]["fx-f64"]: [16 17 18 19 20]
+}
+
+func ExampleFrame_fromArrays() {
+
```
kortschak: Delete blank line.

dframe/dframe_example_test.go
```diff
@@ -0,0 +1,343 @@
...
+	// Output:
+	// cols: [f1-i32 f2-f64]
+	// cols: [f2-f64 fx-f64]
+	// rec[0]["f2-f64"]: [4 5 6 7 8]
+	// rec[0]["fx-f64"]: [4 5 6 7 8]
+}
+
+func ExampleFrame_fromCols() {
+
```
kortschak: Delete blank line.

dframe/dframe_example_test.go
```diff
@@ -0,0 +1,343 @@
...
+	// Output:
+	// cols: [f1-i32 f2-f64]
+	// cols: [f2-f64 fx-f64]
+	// rec[0]["f2-f64"]: [4 5 6 7 8]
+	// rec[0]["fx-f64"]: [4 5 6 7 8]
+}
+
+func ExampleFrame_fromFrame() {
+
```
kortschak: Delete blank line.

dframe/dframe.go
```diff
@@ -9,5 +9,620 @@
...
+	if df.rows < 0 {
+		switch len(df.cols) {
+		case 0:
+			df.rows = 0
+		default:
+			df.rows = int64(df.cols[0].Len())
+		}
+	}
+
+	{
```
kortschak: What is this block for?

dframe/dframe.go
```diff
@@ -9,5 +9,620 @@
...
+	err    error // first error encountered
+	mem    memory.Allocator
+	schema *arrow.Schema
+
+	cols []array.Column
+	rows int64
+}
+
+// Dict is a map of string to array of data.
+type Dict map[string]interface{}
```
kortschak: Is this needed?

dframe/dfmat/dfmat.go
```diff
@@ -0,0 +1,142 @@
...
+func FromVector(vec mat.Vector, opts ...Option) *dframe.Frame {
+	var (
+		mem = memory.NewGoAllocator()
+		bld = array.NewFloat64Builder(mem)
+
+		rows   = vec.Len()
+		arrs   = make([]array.Interface, 1)
+		fields = make([]arrow.Field, 1)
+	)
+	defer bld.Release()
```
kortschak: Same here. I think just break this var decl blocks up.

dframe/dfmat/dfmat.go
```diff
@@ -0,0 +1,142 @@
...
+
+// WithNames configures a dframe.Frame with the provided set of column names.
+func WithNames(names ...string) Option {
+	return func(c *config) {
+		c.names = make([]string, len(names))
+		copy(c.names, names)
+	}
+}
+
+type config struct {
```
kortschak: Could this be omitted and just work on a `[]string`?

btracey:
Dataframes are clearly important for the Go ecosystem. However, I do wonder if Gonum is really the right place for this package. Gonum is nice in that it is self-contained, relying only on the standard library and `exp/rand`. This builds off several core pieces outside of what Go and Gonum provide, and thus would complicate the story of what Gonum "is". Beyond that, it seems to me the lessons of history point to keeping things contained and independent. The Go team has expressed lament that the standard library is as big as it is. The python ecosystem is just that, an ecosystem. Scipy, Sci-kit learn, pandas are all independent packages. Plotting in python has seen a number of different attempts by different groups and make the story exciting. We even see this story within the scientific Go ecosystem. You cite two previous versions of dataframe packages, and Gorgonia has been very successful, and it while it is outside Gonum it can still work with Gonum.

In a different repository there's a lot more freedom, even if it's many of the same people working to develop the codebase. I think the small set of dependencies is important to Gonum, and there are many API consistency considerations, especially involving memory allocation and the ability to make code run in parallel. A package outside of Gonum can easily break these constraints where reasonable to satisfy that package's goals. One idea for a name would be `kugos` which is the word for "bear" in Cebuano (and has an open github address), another idea would be "makwa" (though not open) which means bear in Algonquin.

dframe/README.md
sbinet: it's a bet.
I am betting [Arrow](https://arrow.apache.org) will be a major player in both data science and in many other "big data" areas.
I am betting the Arrow standard (_ie:_ the standardized memory layout of Arrow arrays) will be used in many workloads and data analysis pipelines.

You're right that I haven't properly motivated that bet and that I've just "name dropped" a few features of Arrow: zero-copy mechanism, interoperability with other languages, future (?) interoperability with other big data framework.

I'll address that shortly.

dframe/README.md
sbinet: do you mean the name of the method (`ColumnNames`) is discordant or is the comment?
if the former, what about `Names()` ?

dframe/README.md
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+```
+
+```go
+// Open opens an already existing Frame using the provided driver technology,
+// located at the provided source.
+//
+// Possible drivers: hdf5, npyio, csv, json, hdfs, spark, sql, ...
+func Open(drv, src string) (*Frame, error) { ... }
```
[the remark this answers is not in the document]
sbinet: I am thinking about having a `dfcsv` package where some appropriate function would take the `io.Reader`:

```go
package dfcsv // import "gonum.org/v1/exp/dframe/dfcsv"

func Read(r io.Reader, opts ...Option) (*Frame, error) { ... } // or name it Open?
```

and some `database/sql`-like driver registration code so one could write:
```go
df, err := dframe.Open("csv", "file.csv")
```

dframe/dfmat/dfmat.go
sbinet: I haven't wired up all the things but I was also thinking about being able to allocate a `dframe.Frame` on a GPGPU backend.
this would mean adding another field to `config`:
```go
type config struct {
    mem   memory.Allocator
    names []string
}
```

dframe/dframe.go
sbinet: it makes for a nicer looking example.
compare:
```go
df, err := dframe.FromMem(map[string]interface{}{"col1": arr1, "col2": arr2})
df, err := dframe.FromMem(dframe.Dict{"col1": arr1, "col2": arr2})
```

I was also thinking about adding something like the `plot.Plot.Add` method:
```go
func NewDict(args ...interface{}) Dict {
    d := make(Dict)
    for i := 0; i < len(args); i += 2 {
        k := args[i].(string)
        v := args[i+1]
        d[k] = v
    }
    // check lengths of slices match.
    return d
}
```
this could be removed, though.

dframe/dframe.go
sbinet: copy-pasta...

dframe/README.md
kortschak: It's a discordant approach. One method returns the ith column, the other method returns all column names. `ColumnNames` is also a potential second source of truth for `NumCols`.

I'd suggest `Name(int) string`.

dframe/dframe.go
kortschak: I wouldn't add the `NewDict` function. It adds nothing beyond `dframe.Dict{...`.

dframe/dfmat/dfmat.go
kortschak: OK

sbinet:
@btracey 
> I think the small set of dependencies is important to Gonum

keeping the amount of dependencies to a minimum is good practice.
but I wouldn't use it to prevent `dframe` to find Gonum as a home.

see:
```sh
$> function show-deps() {
for pkg in $(go list ./...); do 
        go list -f "{{range .Deps}}
{{.}}
{{- end}}" $pkg | grep  "\." | grep -v "gonum.org"; 
   done | sort | uniq
}

$> cd $GOPATH/src/gonum.org/v1/gonum
$> show-deps
golang.org/x/exp/rand
golang.org/x/tools/container/intsets

$> cd ../plot
$> show-deps
github.com/ajstarks/svgo
github.com/golang/freetype
github.com/golang/freetype/raster
github.com/golang/freetype/truetype
github.com/jung-kurt/gofpdf
github.com/llgcode/draw2d
github.com/llgcode/draw2d/draw2dbase
github.com/llgcode/draw2d/draw2dimg
golang.org/x/image/draw
golang.org/x/image/font
golang.org/x/image/math/f64
golang.org/x/image/math/fixed
golang.org/x/image/tiff
golang.org/x/image/tiff/lzw
rsc.io/pdf

$> cd ../exp/dframe
$> show-deps
github.com/apache/arrow/go/arrow
github.com/apache/arrow/go/arrow/array
github.com/apache/arrow/go/arrow/internal/bitutil
github.com/apache/arrow/go/arrow/internal/cpu
github.com/apache/arrow/go/arrow/internal/debug
github.com/apache/arrow/go/arrow/memory
github.com/pkg/errors
```

my reasoning for trying to have `dframe` under the Gonum umbrella is:
- to make it adhere to the high standards of Gonum
- to make it as an entry point for people to hopefully then diffuse in other Gonum repos
- to somewhat gather a critical mass of Gonum developers.

right now, under Gonum, we have the following active repos:
- `gonum.org/v1/exp`
- `gonum.org/v1/gonum`
- `gonum.org/v1/hdf5`
- `gonum.org/v1/netlib`
- `gonum.org/v1/plot`
- `gonum.org/v1/tools`

hardly a super crowded place :)

that said, if many Gonum devs feel like `dframe` has no business to live under `gonum.org`, well, I'll try to find another home for it.

dframe/dframe.go
sbinet: ok.

dframe/README.md
sbinet: done.

dframe/README.md
```diff
@@ -0,0 +1,184 @@
...
+However, introducing a `dframe.Tx` transaction could tackle the memory leak.
+One can achieve the above goal if one only allows modifications of the underlying `Frame` through a transaction, where all operations are applied to a single temporary `Frame`:
+
+```go
+// Exec runs the provided function inside an atomic read/write transaction,
+// applied on this Frame.
+func (df *Frame) Exec(f func(tx *Tx) error) error { ... }
+
+func example(df *dframe.Frame) {
+	err := df.Exec(func(tx *dframe.Tx) error {
```
tobgu: The need for reference counting in Arrow will probably incur some mental overhead and be a source of bugs since it's different from how memory is handled in Go normally. This would probably be a bigger issue in the case of an immutable dataframe than a mutable but it will always be present (right?).

I like the idea of a transaction, chained calls or not, immutable or not. Your suggestion would probably give the greatest flexibility, at the cost of some noise.

This is an alternative which removes some noise (function declaration and return) in client code at the cost of sligthly less flexibility.

```go
// Client code
df.Exec(dframe.Slice(0, 10), dframe.Select("col1", "col2"), dframe.Apply("col1 + col2"))

// In package dframe
type ExecFunc func(df *Frame)

func(df *Frame) Exec(ff... ExecFunc) {
	/* apply functions, manage reference counts, etc */ 
}

func(df *Frame) doSelect(cols... string) {
	// Select columns
}

func Select(cols... string) ExecFunc {
	return func(df *Frame) {
                // Either manipulate df directly, or, as below, call package private method on df to do the real work.
		df.doSelect(cols...)
	}
}

func Slice(cols... string) ExecFunc {
	// Similar to above
}
```

sbinet: > The need for reference counting in Arrow will probably incur some mental overhead and be a source of bugs

yeah... I tried at some point to get rid of it in Go Arrow, but I didn't find a way to support GPGPU (or any alternate memory backing store) and drop this ref-counting thing.

hiding the ref-counting behind the `dframe.Tx` transaction was my way to handle this (so one gets only 1 `dframe.Frame` that needs to be `defer`ed with `Release()`, like the usual `os.File.Close`. This greatly reduces the amount of possible code paths that can go wrong.)

> This is an alternative which removes some noise (function declaration and return) in client code at the cost of sligthly less flexibility.

that's an interesting alternative. thanks for bringing it up!
it indeed reduces the amount of user-code "noise".
it creates and returns one closure/function per operation, compared to one big closure/function.
I must say I don't know whether that really matters performance wise (CPU, Mem) though...

tobgu: I would assume it should not matter performance wise for any but the smallest dataframes. But one should not assume things. :-)

nickpoorman:
Wow this looks incredibly promising so far!

dframe/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+Or, without a "chained methods" API:
+
+```go
+func example(df *dframe.Frame) {
+	err := df.Exec(func(tx *dframe.Tx) error {
+		tx.Slice(0, 10)
+		tx.Select("col1", "col2")
+		tx.Apply("col1 + col2")
```
nickpoorman: I'm curious as to your thoughts on how the implementation for this should be approached.

dframe/dframe.go
```diff
@@ -9,5 +9,619 @@
...
+	return df.rows
+}
+
+// NumCols returns the number of columns of this Frame.
+func (df *Frame) NumCols() int64 {
+	return int64(len(df.cols))
+}
+
+// Column returns the i-th column of this Frame.
+func (df *Frame) Column(i int) *array.Column {
```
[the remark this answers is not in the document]
sbinet: you're right.

I've always been a bit split b/w using "natural" primitive type (such as `int`) and also being able to use and follow the arrow standard (ie: `int64`).

I put `Column(int)` because I'd somehow expected the columns would be stored as `[]someType`...

perhaps one should just have `NumCols() int`.

sbinet:
On Sat, Apr 13, 2019 at 7:54 PM Nick Poorman <notifications@github.com>
wrote:

> *@nickpoorman* commented on this pull request.
> ------------------------------
>
> In dframe/README.md
> <https://github.com/gonum/exp/pull/19#discussion_r275126626>:
>
> > +		return nil
> +	})
> +	if err != nil {
> +		log.Fatal(err)
> +	}
> +}
> +```
> +
> +Or, without a "chained methods" API:
> +
> +```go
> +func example(df *dframe.Frame) {
> +	err := df.Exec(func(tx *dframe.Tx) error {
> +		tx.Slice(0, 10)
> +		tx.Select("col1", "col2")
> +		tx.Apply("col1 + col2")
>
> I'm curious as to your thoughts on how the implementation for this should
> be approached.
>

my hand-waving reply would: an SQL-like engine to interpret the expression
and build some kind of AST that would be executed.
but that's as fuzzy as it sounds :)

-s

sbinet:
closing of old age.

PR #21, opened 2019-01-12
linsolve: update for change in mat.DiagDense API

### LINKED ISSUE #20: linsolve: test fails with undefined: mat.NewDiagonal
running `go test` in travis fails (locally too) with:
```
$> go test ./...
# gonum.org/v1/exp/linsolve [gonum.org/v1/exp/linsolve.test]
linsolve/linsolve_test.go:154:7: undefined: mat.NewDiagonal
```

kortschak:
Please take a look.

Fixes #20.

### DIFF
--- a/linsolve/linsolve_test.go
+++ b/linsolve/linsolve_test.go
@@ -151,7 +151,7 @@ func newRandomDiagonal(n int, rnd *rand.Rand) testCase {
 		a[i] = 1 + 10*rnd.Float64()
 		diag[i] = a[i]
 	}
-	A := mat.NewDiagonal(n, a)
+	A := mat.NewDiagDense(n, a)
 	// Generate the right-hand side.
 	b := make([]float64, n)
 	for i := range b {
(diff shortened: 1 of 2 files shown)


linsolve/linsolve_test.go
```diff
@@ -366,7 +366,7 @@ func newGreenbaum73(nx, ny int, rnd *rand.Rand) testCase {
 		negOne, negOne, // - ∂_x ∂_x u - ∂_y ∂_y u
 		func(x, _ float64) float64 { return 40 * x }, // 40 * x * ∂_x u
 		func(_, y float64) float64 { return 40 * y }, // 40 * y * ∂_y u
-		constant(-100),                               // -100 * u
+		constant(-100), // -100 * u
```
kortschak: This change in due, I think, to a change in gofmt tuning.

vladimir-ch: I think it depends on which formatting program is used. I've seen it change between goimports or goreturns, IIRC.

kortschak: Yeah, maybe, but there was a recent change to the line length comparison calculations recently.

PR #24, opened 2019-01-21
linsolve: API changes around Context and Method.Init

vladimir-ch:
The changes are (hopefully well) explained in the individual commit messages.

I'm quite happy with these changes, I think they make much sense, simplify the code and clarify the purpose of the fields of `Context`.

I think these should be the last major API changes.

### DIFF
--- a/linsolve/bicg.go
+++ b/linsolve/bicg.go
@@ -54,7 +54,7 @@ func (b *BiCG) Init(dim int) {
 //  MulVec|Trans
 //  PreconSolve
 //  PreconSolve|Trans
-//  CheckResidual
+//  CheckResidualNorm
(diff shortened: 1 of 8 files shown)


btracey:
This is nice. LGTM. The only thing is possibly adding some documentation that Dst will only be modified at the appropriate times? These methods use Dst as a temporary workspace and it's not clear that's kosher.

vladimir-ch:
I'm not sure if this is what you meant but I added a comment to `Context.Dst` and made the use of `Dst` as a temporary workspace more local so that if we decide against it, it will be a one-line change. I personally think it's fine, it happens only during the initialization and I just don't want to allocate a new slice for this.

PTAL

PR #27, opened 2019-02-18
layout/barneshut: new package for calculating Barnes-Hut force approximations

kortschak:
WIP

Build failures are due to issues with coveralls.io lemurheavy/coveralls-public#1264.

### DIFF
--- a/layout/barneshut/barneshut2.go
+++ b/layout/barneshut/barneshut2.go
@@ -16,27 +16,27 @@ const (
 	nw
 )
 
-// Point2 is a 2D point.
-type Point2 struct {
+// Vector2 is a 2D vector.
+type Vector2 struct {
 	X, Y float64
 }
 
 // Add returns the vector sum of p and q.
-func (p Point2) Add(q Point2) Point2 {
+func (p Vector2) Add(q Vector2) Vector2 {
 	p.X += q.X
 	p.Y += q.Y
 	return p
 }
 
 // Sub returns the vector sum of p and -q.
-func (p Point2) Sub(q Point2) Point2 {
+func (p Vector2) Sub(q Vector2) Vector2 {
 	p.X -= q.X
 	p.Y -= q.Y
 	return p
 }
 
 // Scale returns the vector p scaled by f.
-func (p Point2) Scale(f float64) Point2 {
+func (p Vector2) Scale(f float64) Vector2 {
 	p.X *= f
 	p.Y *= f
 	return p
@@ -44,12 +44,12 @@ func (p Point2) Scale(f float64) Point2 {
 
 // Box2 is a 2D bounding box.
 type Box2 struct {
-	Min, Max Point2
+	Min, Max Vector2
 }
 
 // quadrant returns which quadrant of b that p should be placed in.
 func (b Box2) quadrant(p Particle2) int {
-	center := Point2{
+	center := Vector2{
 		X: (b.Min.X + b.Max.X) / 2,
 		Y: (b.Min.Y + b.Max.Y) / 2,
 	}
@@ -95,7 +95,7 @@ func (b Box2) split(dir int) Box2 {
 
 // Particle2 is a particle in a plane.
 type Particle2 interface {
-	Coord2() Point2
+	Coord2() Vector2
 	Mass() float64
 }
 
@@ -107,16 +107,16 @@ type Particle2 interface {
 // compared. Force2 may be passed nil for p2 when the Barnes-Hut approximation
 // is being used. A nil p2 indicates that the second mass center is an
 // aggregate.
-type Force2 func(p1, p2 Particle2, m1, m2 float64, v Point2) Point2
+type Force2 func(p1, p2 Particle2, m1, m2 float64, v Vector2) Vector2
 
 // Gravity2 returns a vector force on m1 by m2, equal to (m1⋅m2)/‖v‖²
 // in the directions of v. Gravity2 ignores the identity of the interacting
 // particles and returns a zero vector when the two particles are
 // coincident, but performs no other sanity checks.
-func Gravity2(_, _ Particle2, m1, m2 float64, v Point2) Point2 {
+func Gravity2(_, _ Particle2, m1, m2 float64, v Vector2) Vector2 {
 	d2 := v.X*v.X + v.Y*v.Y
 	if d2 == 0 {
-		return Point2{}
+		return Vector2{}
 	}
 	return v.Scale((m1 * m2) / (d2 * math.Sqrt(d2)))
 }
@@ -184,7 +184,7 @@ func (q *Plane) Reset() {
 // interaction is with a non-aggregate mass center, otherwise p2 will be nil.
 //
 // It is safe to call ForceOn concurrently.
-func (q *Plane) ForceOn(p Particle2, theta float64, f Force2) (vector Point2) {
+func (q *Plane) ForceOn(p Particle2, theta float64, f Force2) (force Vector2) {
 	var empty tile
 	if theta > 0 && q.root != empty {
 		return q.root.forceOn(p, p.Coord2(), p.Mass(), theta, f)
@@ -192,7 +192,7 @@ func (q *Plane) ForceOn(p Particle2, theta float64, f Force2) (vector Point2) {
 
 	// For the degenerate case, just iterate over the
 	// slice of particles rather than walking the tree.
-	var v Point2
+	var v Vector2
 	m := p.Mass()
 	pv := p.Coord2()
 	for _, e := range q.Particles {
@@ -209,7 +209,7 @@ type tile struct {
 
 	nodes [4]*tile
 
-	center Point2
+	center Vector2
 	mass   float64
 }
 
@@ -230,7 +230,7 @@ func (t *tile) insert(p Particle2) {
 	t.passDown(p)
 	t.passDown(t.particle)
 	t.particle = nil
-	t.center = Point2{}
+	t.center = Vector2{}
 	t.mass = 0
 }
 
@@ -243,7 +243,7 @@ func (t *tile) passDown(p Particle2) {
 }
 
 // summarize updates node masses and centers of mass.
-func (t *tile) summarize() (center Point2, mass float64) {
+func (t *tile) summarize() (center Vector2, mass float64) {
 	for _, d := range &t.nodes {
 		if d == nil {
 			continue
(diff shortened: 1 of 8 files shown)


kortschak:
The figure for the aborjs planar example taking into accound the bounds changes is here.

example-space-orig

kortschak:
Please take a look.

I think the overall package and testing are OK, but I'm not sure of whether to provide the compact construction as a option to the `New` functions. This would require digging the old `insert` and `forcesOnMassAt` methods as the compact version methods, and adding a `compact bool` field to `Plane` and `Volume` to choose which forces method to call.

layout/barneshut/barneshut2.go
```diff
@@ -0,0 +1,250 @@
...
+	dir := t.bounds.quadrant(p)
+	if t.nodes[dir] == nil {
+		t.nodes[dir] = &tile{bounds: t.bounds.split(dir)}
+	}
+	t.nodes[dir].insert(p)
+}
+
+// summarize updates node masses and centers of mass.
+func (t *tile) summarize() (center Point2, mass float64) {
+	for _, d := range &t.nodes {
```
sbinet: just a quick drive-by comment:
why `&t.nodes` ? (and not just `t.nodes`)

(ditto for `bucket.summarize()` )

kortschak: The `nodes` is an array, so this avoids a copy.

sbinet: Is range really copying the array?
I knew that when one did:
```go
for i,v := range vs
```
The `v` value was being copied.
But vs? Didn't know... Are you sure?

kortschak: https://play.golang.org/p/SBTchSt2rU5

sbinet: thanks. that's subtle.

kortschak:
*ping*

layout/barneshut/barneshut2.go
```diff
@@ -0,0 +1,278 @@
...
+
+// Force2 is a force modeling function for interactions between p1 and p2,
+// m1 is the mass of p1 and m2 of p2. The vector v is the vector from p1 to
+// p2. The returned value is the force vector acting on p1.
+//
+// In models where the identity of particles must be known, p1 and p2 may be
+// compared. Force2 may be passed nil for p2 when the Barnes-Hut approximation
+// is being used. A nil p2 indicates that the second mass center is an
+// aggregate.
+type Force2 func(p1, p2 Particle2, m1, m2 float64, v Point2) Point2
```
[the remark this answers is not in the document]
kortschak: Yes. In the optimising case, the second particle is not real and so is nil.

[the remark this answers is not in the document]
kortschak: Looking now, I suspect that I have that wrong, since there is no way that `p2` is ever nil AFAICS..

kortschak:
*ping*

layout/barneshut/barneshut2_test.go
```diff
@@ -0,0 +1,523 @@
...
+
+func TestPlane(t *testing.T) {
+	const tol = 1e-15
+
+	for _, test := range planeTests {
+		var particles []Particle2
+		if test.particles != nil {
+			particles = make([]Particle2, len(test.particles))
+		}
+		for i, p := range test.particles {
```
sbinet: couldn't we use `copy(particles, test.particles)` here?

layout/barneshut/barneshut3_test.go
```diff
@@ -0,0 +1,520 @@
...
+
+func TestVolume(t *testing.T) {
+	const tol = 1e-15
+
+	for _, test := range volumeTests {
+		var particles []Particle3
+		if test.particles != nil {
+			particles = make([]Particle3, len(test.particles))
+		}
+		for i, p := range test.particles {
```
sbinet: ditto.

layout/barneshut/galaxy_example_test.go
```diff
@@ -0,0 +1,77 @@
...
+		for j, s := range stars {
+			vectors[j] = plane.ForceOn(s, theta, barneshut.Gravity2).Scale(G)
+		}
+
+		// Update positions.
+		for j, s := range stars {
+			s.move(vectors[j])
+		}
+
+		// Rendering stars is left as an exercise for
```
sbinet: pun aside, that would be a great PR plot :)
(or, even, a nice animated GIF)

layout/barneshut/barneshut2_test.go
kortschak: No, they're not the same type: `[]Particle2` and `[]particle2`. One has an interface element and the other a concrete element.

layout/barneshut/barneshut3_test.go
kortschak: Same

kortschak:
Do you have any views on the slightly slower by considerably less space-costly packing?

When I merge this, I'm going to leave the full history so that we can recover that if needed. In the long run I want to merge this into gonum/gonum for two things: graph layouting (hence the name of the branch) and BH t-SNE. I don't see a good location for it in that tree though. Any suggestions?

sbinet:
> Do you have any views on the slightly slower b[ut] considerably less space-costly packing?

I am fine with that.

> I don't see a good location for it in that tree though. Any suggestions?

perhaps `gonum/graph/layout` ?

kortschak:
I will initially use the canonical implementation, but might add back the space-compact algorithm later (with an option to use it rather than by default).

I had initially thought that it would go in graph, but BH-tSNE doesn't fit there. I'm thinking a layout top-level.

sbinet:
> I had initially thought that it would go in graph, but BH-tSNE doesn't fit there. I'm thinking a "gonum/layout" or "gonum/spatial" top-level.

SGTM.
`scipy` has `spatial`:
```python
>>> import scipy.spatial as spa
>>> dir(spa)
['ConvexHull',
 'Delaunay',
 'HalfspaceIntersection',
 'KDTree',
 'Rectangle',
 'SphericalVoronoi',
 'Voronoi',
 'cKDTree',
 'ckdtree',
 'convex_hull_plot_2d',
 'delaunay_plot_2d',
 'distance',
 'distance_matrix',
 'division',
 'kdtree',
 'minkowski_distance',
 'minkowski_distance_p',
 'procrustes',
 'qhull',
 'transform',
 'tsearch',
 'voronoi_plot_2d']
```

kortschak:
Excellent.

I'd like at least one other reviewer on the implementation before I merge this here. Then I can propose it there.

@btracey @vladimir-ch Do either of you have time?

vladimir-ch:
Not earlier than Tuesday next week.

kortschak:
This can wait.

layout/barneshut/barneshut2.go
```diff
@@ -0,0 +1,278 @@
...
+)
+
+const (
+	ne = iota
+	se
+	sw
+	nw
+)
+
+// Point2 is a 2D point.
```
vladimir-ch: This type is used mostly for force calculations, so I think it would be clearer to call it `Vec2` or `Vector2`. Even where it is used for particle position it can be considered as a position _vector_.

Also, I think it would be nice to have `r2` and `r3` packages as in the `geo/r2,r3` packages. Then the users of this package would implement their particles against `r2.Vec` or `r3.Vec` types instead of (longer and less natural) `barneshut.Point2,3`. What do you think?

layout/barneshut/galaxy_example_test.go
```diff
@@ -0,0 +1,77 @@
...
+		p[i] = s
+	}
+	vectors := make([]barneshut.Point2, len(stars))
+
+	// Make a plane to calculate approximate forces
+	plane := barneshut.Plane{Particles: p}
+
+	// Run a simulation for 100 updates.
+	for i := 0; i < 1000; i++ {
+		// Build the data structure, For small system
```
vladimir-ch: Dot instead of comma, and plural in `systems` (or an `a`).

layout/barneshut/barneshut2.go
```diff
@@ -0,0 +1,278 @@
...
+		X: (b.Min.X + b.Max.X) / 2,
+		Y: (b.Min.Y + b.Max.Y) / 2,
+	}
+	c := p.Coord2()
+	if checkBounds && (c.X < b.Min.X || b.Max.X < c.X || c.Y < b.Min.Y || b.Max.Y < c.Y) {
+		panic(fmt.Sprintf("p out of range %+v: %#v", b, p))
+	}
+	if c.X < center.X {
+		if c.Y < center.Y {
+			return nw
```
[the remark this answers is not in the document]
kortschak: It is clock oriented sectors, counting clockwise. The coordinates are going to be used in graphics (primarily), so the coords are graphics display orientation.

layout/barneshut/barneshut2.go
```diff
@@ -0,0 +1,278 @@
...
+	p.Y *= f
+	return p
+}
+
+// Box2 is a 2D bounding box.
+type Box2 struct {
+	Min, Max Point2
+}
+
+// quadrant returns which quadrant of b that p should be placed in.
```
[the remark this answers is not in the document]
kortschak: It is. It could be rewritten to avoid the final preposition, but it's not necessary.

layout/barneshut/galaxy_example_test.go
kortschak: Fixed.

layout/barneshut/barneshut2.go
kortschak: The vector change sounds reasonable.

For the `r2`/`r3` changes where would these be placed? @sbinet suggested that this end up in gonum/spatial, so gonum/spatial/{r2,r3}? I would do that when I bring this over there rather than in this PR.

layout/barneshut/barneshut2.go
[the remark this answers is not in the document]
vladimir-ch: Thanks.

layout/barneshut/barneshut2.go
vladimir-ch: All SGTM.

layout/barneshut/barneshut2.go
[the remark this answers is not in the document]
vladimir-ch: That's what I suspected.

kortschak:
Do you have any views about the alternative non-leaf tree packing? It's slower because it performs more force calculations, but it packs more efficiently because more internal node slots are used. It's marginally more acurate because, again more calculations are performed. I'll merge it as is but we can revisit this later if it's seen as worth following.

PR #29, opened 2019-02-20
plotter/rings: update for changes in vgimg back-end

kortschak:
@sbinet @fogleman

The gg changes have some unfortunate impacts on arc drawing (which we don't test elsewhere other than here).

A failure that shows starkly what is happening is this [one](https://github.com/gonum/exp/compare/rings/gg?expand=1&short_path=f483f55#diff-f483f5559a0d406cdd0936e5cffcb6b8). Just for clirification, the deleted image was generated via gonum/plot with the llgo package and the new image is the same generated via the gg package. Images generated with vgsvg match the deleted images.

Do you have any ideas about what is happening?

I've looked at the relevant gg `ctx` methods ([here](https://github.com/fogleman/gg/blob/9db508d34a2da0d0a3244cf73bb605e2c00c6229/context.go#L579-L603) and [here](https://github.com/fogleman/gg/blob/9db508d34a2da0d0a3244cf73bb605e2c00c6229/context.go#L611-L613)) and there's no docs explaining the parameters or tests for the behaviour, so I'm not sure where to start.


kortschak:
I've figured this out. It's due to a difference in arc description concepts. See gonum/plot#527.

kortschak:
@sbinet PTAL

go.mod
```diff
@@ -0,0 +1,10 @@
+module gonum.org/v1/exp
+
+go 1.12
```
sbinet: perhaps remove this for the time being?

PR #33, opened 2019-03-31
github,travis: update configs and templates

kortschak:
Please take a look.


kortschak:
Yes, I did. I was surprised there was so much stuff there too.

PR #34, opened 2019-04-06
travis: add macOS and Windows to the build matrix

kortschak:
run-parts is reimplemented in simple form since macOS and Windows do not have it. Windows excludes go1.11 and go1.10 since these do no proceed, although they appear to exist. The master branch for Go is excluded because that does not exist for Windows.

This adds significant time to our testing (3x when the later two Go versions are working). This will be significant in the context of gonum/gonum. To be perfectly honest, I'm not entirely sure why we test on Windows(/macOS) since we don't have any OS-specific build tags; everything is depending on Go being cross-platform which is tested elsewhere (golang.org).

To reduce the cost, I have not done any coverage or formatting checks for the commercial platforms and because of the absence/differences/inadequacies in tooling, I have not done code gen checks.

Please take a look.


.travis.yml
```diff
@@ -15,36 +20,22 @@ env:
...
- - go test -v ./...
- - go test -a -tags bounds -v ./...
- - go test -a -tags noasm -v ./...
- - go test -a -tags appengine -v ./...
- - if [[ $TRAVIS_SECURE_ENV_VARS = "true" ]]; then bash ./.travis/test-coverage.sh; fi
- - ${TRAVIS_BUILD_DIR}/.travis/check-imports.sh
- # This is run last since it alters the tree.
- - ${TRAVIS_BUILD_DIR}/.travis/check-generate.sh
-
+ - source ${TRAVIS_BUILD_DIR}/.travis/run-parts ${TRAVIS_BUILD_DIR}/.travis/run.d/${TRAVIS_OS_NAME}
```
sbinet: do we really need to source that script?

kortschak: I can try without. I'm not sure exactly what travis do with their scripts.

kortschak: OK. That works for me. It also means we can `exit 1` in the `run-parts` loop for failures if we want.

kortschak:
Should note that the test output doesn't end up in the logs for windows. Maybe this is what they mean by black box testing.

kortschak:
>my take is that making sure we build+test for `{darwin,windows} x 1-variant` ensures (well, helps at least) we don't paint ourselves into a linux corner.

Surely we can get this from having a vet for acceptable build tags and file suffixes (which would not include any of the operating system tags). This obviously precludes any operating system differentiation, but then I don't see any real need at our layer of the stack.

What I do see testing as touching on marketing; we demonstrate that we work on multiple platforms.

PR #37, opened 2019-07-30
yaegi: add symbol lookup table for yaegi interpreter

kortschak:
I haven't played with this at all, but it looks like it's necessary for gonum to be used in the interpreter when it's embedded.

The generated code doesn't conform to what we need for the core repo, but it's fine here.

Please take a look.


sbinet:
Looks good.
I was bothered by the missing LICENSE blurb (and the ordering of some comments), so I've created https://github.com/containous/yaegi/issues/310.

perhaps wait for the associated PR to be merged before letting this one go in?

kortschak:
Yeah, I'd be happy to have that go in first. I was considering sending a PR, but I didn't feel comfortable going into someone else's project and rearranging things without having contributed more substantially.

kortschak:
There's also the orphan comments in the `Symbols` assignments that could be fixed with a simple conditional prior to the comment, for example [here](https://github.com/gonum/exp/pull/37/files#diff-1c1a25df3cb48201d75925bbc563a652R17).

sbinet:
this has been addressed in my latest PR revision.

kortschak:
PTAL

The warning doesn't mention the license parameter, which is troubling. Also, it's not possible to use go run, or to ensure that `goexports` is installed in the bash script, presumably due to modules. If either of those things are done, you see many of the following:
```
2019/08/05 20:32:43 go/build: importGo -license: exit status 2
flag provided but not defined: -license
usage: go list [-f format] [-json] [-m] [list flags] [build flags] [packages]
Run 'go help list' for details.
```

kortschak:
PTAL

kortschak:
Yeah, I have run `go mod tidy` here.

PR #40, opened 2019-11-02
linsolve: add package comment and examples

vladimir-ch:
It's not perfect but it's either now or never. After this PR is merged, the base of the package is finished and so I'd like to finally move this to the main Gonum repository.

Please take a look.

### DIFF
--- a/linsolve/bicg.go
+++ b/linsolve/bicg.go
@@ -10,10 +10,12 @@ import (
 	"gonum.org/v1/gonum/floats"
 )
 
-// BiCG implements the Bi-Conjugate Gradient method with
-// preconditioning for solving systems of linear equations
-//  A*x = b,
-// where A is a square, possibly nonsymmetric matrix.
+// BiCG implements the Bi-Conjugate Gradient method with preconditioning for
+// solving systems of linear equations
+//  A * x = b,
+// where A is a nonsymmetric, nonsingular matrix. It uses limited memory storage
+// but the convergence may be irregular and the method may break down. BiCG
+// requires a multiplication with A and A^T at each iteration.
 //
 // References:
 //  - Barrett, R. et al. (1994). Section 2.3.5 BiConjugate Gradient (BiCG).
--- a/linsolve/bicg.go
+++ b/linsolve/bicg.go
@@ -15,7 +15,8 @@ import (
 //  A * x = b,
 // where A is a nonsymmetric, nonsingular matrix. It uses limited memory storage
 // but the convergence may be irregular and the method may break down. BiCG
-// requires a multiplication with A and A^T at each iteration.
+// requires a multiplication with A and Aᵀ at each iteration. BiCGStab is a
+// related method that does not use Aᵀ.
 //
 // References:
 //  - Barrett, R. et al. (1994). Section 2.3.5 BiConjugate Gradient (BiCG).
@@ -99,7 +100,7 @@ func (b *BiCG) Iterate(ctx *Context) (Operation, error) {
 	case 4:
 		// z is overwritten and reused.
 		copy(b.z, ctx.Dst)
-		// Compute A^T * pt.
+		// Compute Aᵀ * pt.
 		copy(ctx.Src, b.pt)
 		b.resume = 5
 		return MulVec | Trans, nil
--- a/linsolve/cg.go
+++ b/linsolve/cg.go
@@ -8,10 +8,11 @@ import (
 	"gonum.org/v1/gonum/floats"
 )
 
-// CG implements the Conjugate Gradient iterative method with
-// preconditioning for solving systems of linear equations
-//  A*x = b,
-// where A is a symmetric positive definite matrix.
+// CG implements the Conjugate Gradient iterative method with preconditioning
+// for solving systems of linear equations
+//  A * x = b,
+// where A is a symmetric positive definite matrix. It requires minimal memory
+// storage and is the method of choice for symmetric positive definite problem.
 //
 // References:
 //  - Barrett, Richard et al. (1994). Section 2.3.1 Conjugate Gradient Method (CG).
--- a/linsolve/cg.go
+++ b/linsolve/cg.go
@@ -12,7 +12,7 @@ import (
 // for solving systems of linear equations
 //  A * x = b,
 // where A is a symmetric positive definite matrix. It requires minimal memory
-// storage and is the method of choice for symmetric positive definite problem.
+// storage and is a good choice for symmetric positive definite problem.
 //
 // References:
 //  - Barrett, Richard et al. (1994). Section 2.3.1 Conjugate Gradient Method (CG).
--- a/linsolve/cg.go
+++ b/linsolve/cg.go
@@ -1,4 +1,4 @@
-// Copyright ©2016 The Gonum Authors. All rights reserved.
+// Copyright ©2017 The Gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
@@ -12,7 +12,7 @@ import (
 // for solving systems of linear equations
 //  A * x = b,
 // where A is a symmetric positive definite matrix. It requires minimal memory
-// storage and is a good choice for symmetric positive definite problem.
+// storage and is a good choice for symmetric positive definite problems.
 //
 // References:
 //  - Barrett, Richard et al. (1994). Section 2.3.1 Conjugate Gradient Method (CG).
--- a/linsolve/doc.go
+++ b/linsolve/doc.go
@@ -54,9 +54,8 @@ MulVecToer interface
 
 The MulVecToer interface represents the system matrix A in this package. This
 abstracts the details of any particular matrix storage, and allows the user to
-exploit the properties of their particular matrix. All matrix types provided by
-gonum.org/v1/gonum/mat and github.com/james-bowman/sparse packages implement
-this interface.
+exploit the properties of their particular matrix. Matrix types provided by
+github.com/james-bowman/sparse packages implement this interface.
 
 Note that methods in this package have only limited means for checking whether
 the provided MulVecToer represents a matrix that satisfies all assumptions made
--- a/linsolve/doc.go
+++ b/linsolve/doc.go
@@ -14,77 +14,82 @@ A system of linear equations can be written as
 where A is a given n×n non-singular matrix, b is a given n-vector (the
 right-hand side), and x is an unknown n-vector.
 
-Direct methods such as the LU or QR decomposition can find (in the absence of
+Direct methods such as the LU or QR decomposition compute (in the absence of
 roundoff errors) the exact solution after a finite number of steps. For a
 general matrix A they require O(n^3) arithmetic operations which becomes
-infeasible for large problems due to excessive memory and time cost.
+infeasible for large n due to excessive memory and time cost.
 
 Iterative methods, in contrast, generally do not compute the exact solution x.
 Starting from an initial estimate x_0, they instead compute a sequence x_i of
-increasingly accurate approximations to x. When the error of the approximation
-x_i becomes smaller that a prescribed threshold,  x_{i+1} is not computed and
-the iterative process is stopped.
+increasingly accurate approximations to x. This iterative process is stopped
+when the estimated difference between x_i and the true x becomes smaller than a
+prescribed threshold. The number of iterations thus depends on the value of the
+threshold. If the desired threshold is very small, then the iterative methods
+may take as long or longer than a direct method. However, for many problems a
+decent approximation can be found in a small number of steps.
 
-This package provides iterative methods that access the matrix A only only via
-the matrix-vector multiplication. The multiplication requires O(n^2) arithmetic
-operations for a general n×n matrix but can be much cheaper if the matrix is
-sparse (has only a few non-zeros per row) or has some other special properties
-or structure. This is often the case for matrices that arise in practical
-applications. Furthermore, if an iterative method can find a good approximation
-with a number of matrix-vector multiplications that is much lower than n or
-independent of n, it can outperform a direct method for the same problem.
+THe iterative methods implemented in this package do not access the elements of
+A directly, they instead ask for the result of matrix-vector products with A.
+For a general n×n matrix this requires O(n^2) operations, but can be much
+cheaper depending on the structure of the matrix (sparse, banded,
+block-stuctured, etc.). Such structure often arises in practical applications.
+An iterative method can thus be significantly cheaper than a direct method, by
+using a small number of iterations and taking advantage of matrix structure.
 
-This means that iterative methods are most often useful in the following
-situations:
+Iterative methods are most often useful in the following situations:
 
- - the system matrix A is sparse, blocked or has other special structure,
- - the problem size is sufficiently large that a dense factorization of A is too
+ - The system matrix A is sparse, blocked or has other special structure,
+ - The problem size is sufficiently large that a dense factorization of A is
    costly in terms of computer time and/or storage,
- - computing the product of A (or A^T, if necessary) with a vector can be done
-   more efficiently than if A were first formed explicitly as a matrix,
- - it suffices to find only an approximation to the solution.
+ - Computing the product of A (or A^T, if necessary) with a vector can be done
+   efficiently,
+ - An approximate solution is all that is required.
 
 Using linsolve
 
-The two most important elements of the API for solving linear systems are the
-MulVecToer interface and the Iterative function.
+The two most important elements of the API are the MulVecToer interface and the
+Iterative function.
 
-The MulVecToer interface represents the system matrix A in this package. It
-separates the method implementations from the details of any particular matrix
-storage format. All matrix types provided by gonum.org/v1/gonum/mat and
-github.com/james-bowman/sparse packages implement this interface.
+MulVecToer interface
+
+The MulVecToer interface represents the system matrix A in this package. This
+abstracts the details of any particular matrix storage, and allows the user to
+exploit the properties of their particular matrix. All matrix types provided by
+gonum.org/v1/gonum/mat and github.com/james-bowman/sparse packages implement
+this interface.
 
 Note that methods in this package have only limited means for checking whether
 the provided MulVecToer represents a matrix that satisfies all assumptions made
-by the chosen Method, for example whether the matrix is actually symmetric
-positive definite.
+by the chosen Method, for example if the matrix is actually symmetric positive
+definite.
+
+Iterative function
 
 The Iterative function is the entry point to the functionality provided by this
-package. It takes as parameters the matrix A (as a MulVecToer), the right-hand
-side vector b, the iterative method and settings that control the iterative
-process and provide a way for reusing memory.
+package. It takes as parameters the matrix A (via the MulVecToer interface as
+discussed above), the right-hand side vector b, the iterative method and
+settings that control the iterative process and provide a way for reusing
+memory.
 
 Choosing an iterative method
 
 The choice of an iterative method is typically guided by the properties of the
-matrix A including symmetry, definiteness, sparsity, conditioning and block
-structure. This is done to assure convergence of the algorithm for the given
-problem and efficiency of the computation. The methods of choice are well
-understood for symmetric problems in which case their convergence rate (at least
-in exact arithmetics) is determined by spectral properties (eigenvalues) of A.
-The situation is not as clear for problems with non-symmetric matrices and it
-can be difficult to pick the best method for these problems. Any general advice
-for non-symmetric problem is often complemented with a trial-and-error approach.
+matrix A including symmetry, definiteness, sparsity, conditioning, and block
+structure. In general, performance on symmetric matrices is well understood (see
+the references below), with the conjugate gradient method being a good starting
+point. Non-symmetric matrices are much more difficult to assess, where any
+suggestion of a 'best' method is usually accompanied by a recommendation to use
+trial-and-error.
 
 Preconditioning
 
 Preconditioning is a family of techniques that attempt to transform the linear
-system into one that has the same solution but more favorable spectrum. The
-transformation matrix is called a preconditioner. A good preconditioner improves
-the convergence rate of the iterative method, sufficiently to overcome the extra
-cost of constructing and applying the preconditioner. Without a preconditioner
-the iterative method may even fail to converge. In linsolve a preconditioner is
-specified by Settings.PreconSolve.
+system into one that has the same solution but more favorable eigenspectrum. The
+transformation matrix is called a preconditioner. A good preconditioner will
+reduce the number of iterations needed to find a good approximate solution
+(hopefully enough to overcome the cost of applying the preconditioning!), and in
+some cases preconditioning is necessary to get any kind of convergence. In
+linsolve a preconditioner is specified by Settings.PreconSolve.
 
 Implementing Method interface
 
--- a/linsolve/doc.go
+++ b/linsolve/doc.go
@@ -28,7 +28,7 @@ threshold. If the desired threshold is very small, then the iterative methods
 may take as long or longer than a direct method. However, for many problems a
 decent approximation can be found in a small number of steps.
 
-THe iterative methods implemented in this package do not access the elements of
+The iterative methods implemented in this package do not access the elements of
 A directly, they instead ask for the result of matrix-vector products with A.
 For a general n×n matrix this requires O(n^2) operations, but can be much
 cheaper depending on the structure of the matrix (sparse, banded,
@@ -40,7 +40,7 @@ Iterative methods are most often useful in the following situations:
 
  - The system matrix A is sparse, blocked or has other special structure,
  - The problem size is sufficiently large that a dense factorization of A is
-   costly in terms of computer time and/or storage,
+   costly in terms of compute time and/or memory storage,
  - Computing the product of A (or A^T, if necessary) with a vector can be done
    efficiently,
  - An approximate solution is all that is required.
@@ -52,9 +52,9 @@ Iterative function.
 
 MulVecToer interface
 
-The MulVecToer interface represents the system matrix A in this package. This
-abstracts the details of any particular matrix storage, and allows the user to
-exploit the properties of their particular matrix. Matrix types provided by
+The MulVecToer interface represents the system matrix A. This abstracts the
+details of any particular matrix storage, and allows the user to exploit the
+properties of their particular matrix. Matrix types provided by gonum/mat and
 github.com/james-bowman/sparse packages implement this interface.
 
 Note that methods in this package have only limited means for checking whether
--- a/linsolve/doc.go
+++ b/linsolve/doc.go
@@ -3,28 +3,112 @@
 // license that can be found in the LICENSE file.
 
 /*
-Package linsolve provides iterative algorithms for solving linear systems.
+Package linsolve provides iterative methods for solving linear systems.
 
-References:
- - Barrett, Richard et al. (1994). Templates for the Solution of Linear Systems: Building
-   Blocks for Iterative Methods (2nd ed.). Philadelphia, PA: SIAM.
+Background
+
+A system of linear equations can be written as
+
+ A * x = b,
+
+where A is a given n×n non-singular matrix, b is a given n-vector (the
+right-hand side), and x is an unknown n-vector.
+
+Direct methods such as the LU or QR decomposition can find (in the absence of
+roundoff errors) the exact solution after a finite number of steps. For a
+general matrix A they require O(n^3) arithmetic operations which becomes
+infeasible for large problems due to excessive memory and time cost.
+
+Iterative methods, in contrast, generally do not compute the exact solution x.
+Starting from an initial estimate x_0, they instead compute a sequence x_i of
+increasingly accurate approximations to x. When the error of the approximation
+x_i becomes smaller that a prescribed threshold,  x_{i+1} is not computed and
+the iterative process is stopped.
+
+This package provides iterative methods that access the matrix A only only via
+the matrix-vector multiplication. The multiplication requires O(n^2) arithmetic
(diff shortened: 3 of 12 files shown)


linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+	n := len(x)
+	ax := mat.NewVecDense(n, dst)
+	ax.MulVec(ac.a, mat.NewVecDense(n, x))
+}
+
+func (ac *AllenCahnFD) Solution() []float64 {
+	return ac.u
+}
+
+func output(u []float64, L float64, step int) {
```
golangcibot: `output` is unused (from `deadcode`)

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+the iterative method may even fail to converge. In linsolve a preconditioner is
+specified by Settings.PreconSolve.
+
+Implementing Method interface
+
+This package allows external implementations of iterative solvers by means of
+the Method interface. It uses a reverse-communication style of API to
+"outsource" operations such as matrix-vector multiplication, preconditioner
+solve or convergence checks to the caller. The caller performs the commanded
+operation and passes the result again to Method. The matrix A and the right-hand
```
kortschak: Comma bracketing of 'A' and 'b'?

linsolve/example_cg_test.go
```diff
@@ -9,15 +9,20 @@ import (
 	"math"
 
 	"gonum.org/v1/exp/linsolve"
-	"gonum.org/v1/gonum/floats"
 	"gonum.org/v1/gonum/mat"
 )
 
 type system struct {
```
kortschak: Since this is a single file example, maybe add a brief doc comment to `system`.

linsolve/example_cg_test.go
```diff
@@ -9,15 +9,20 @@ import (
...
-	A *mat.SymBandDense
+	a *mat.SymBandDense
 	b []float64
 }
 
+func (sys system) MulVecTo(dst []float64, trans bool, x []float64) {
```
kortschak: Here too; adding a comment that this method provides the reverse communication hook into the method will reinforce the explanation in the package-level documentation.

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
+// Copyright ©2019 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package linsolve_test
```
kortschak: Add in blank line.

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	// Allocate the linear solver and the settings.
+	ac.ls = &linsolve.CG{}
+	ac.lssettings = linsolve.Settings{
+		// Solution from the previous time step will be a good initial estimate.
+		InitX: ac.u,
+		// Store the solution into the existing slice.
+		Dst: ac.u,
+		// Provide context to reduce memory garbage.
```
kortschak:
```
// Provide context to reduce memory allocation and GC pressure.
```

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+			return 0.01 * rnd.NormFloat64()
+		},
+	}
+	ac.Setup(n, L, tau)
+	for i := 1; i <= numSteps; i++ {
+		err := ac.Step()
+		if err != nil {
+			log.Fatal(err)
+		}
+		// output(ac.Solution(), L, i)
```
kortschak: This is the cause of the lint warning above.

linsolve/cg.go
```diff
@@ -8,10 +8,11 @@ import (
...
 
-// CG implements the Conjugate Gradient iterative method with
-// preconditioning for solving systems of linear equations
-//  A*x = b,
-// where A is a symmetric positive definite matrix.
+// CG implements the Conjugate Gradient iterative method with preconditioning
+// for solving systems of linear equations
+//  A * x = b,
+// where A is a symmetric positive definite matrix. It requires minimal memory
+// storage and is the method of choice for symmetric positive definite problem.
```
btracey: a good choice? I'm not sure we want to make such a strong statement as "the method of choice" given how many custom algorithms there are.

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+Background
+
+A system of linear equations can be written as
+
+ A * x = b,
+
+where A is a given n×n non-singular matrix, b is a given n-vector (the
+right-hand side), and x is an unknown n-vector.
+
+Direct methods such as the LU or QR decomposition can find (in the absence of
```
btracey: s/can find/compute, or some other word that's more definitive. "can find" is like an optimization "can find" an optimum, or this package "can find" the true answer. QR is much more solid

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+roundoff errors) the exact solution after a finite number of steps. For a
```
btracey: maybe "fixed" rather than "finite"?

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+general matrix A they require O(n^3) arithmetic operations which becomes
+infeasible for large problems due to excessive memory and time cost.
```
btracey: s/problems/n

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+
+Iterative methods, in contrast, generally do not compute the exact solution x.
+Starting from an initial estimate x_0, they instead compute a sequence x_i of
+increasingly accurate approximations to x. When the error of the approximation
```
btracey: Maybe "This iterative process is stopped when the difference between x_i and the true x becomes smaller than a prescribed threshold"

btracey: Then something like "The number of iterations thus depends on the value of the threshold. If the desired threshold is very small, then the iterative methods may take as long or longer than a direct method. However, for many problems a decent approximation can be found in a small number of steps"

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+x_i becomes smaller that a prescribed threshold,  x_{i+1} is not computed and
+the iterative process is stopped.
+
+This package provides iterative methods that access the matrix A only only via
```
btracey: The iterative methods implemented in this package do not access the elements of A directly, they instead ask for the result of matrix-vector products with A.

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+the matrix-vector multiplication. The multiplication requires O(n^2) arithmetic
```
btracey: For a general nxn matrix this requires O(n^2) operations, but can be much cheaper depending on the structure of the matrix (sparse, banded, block-structured, etc.). Such structure often arises in practical applications.

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+operations for a general n×n matrix but can be much cheaper if the matrix is
+sparse (has only a few non-zeros per row) or has some other special properties
+or structure. This is often the case for matrices that arise in practical
+applications. Furthermore, if an iterative method can find a good approximation
```
btracey: From "Futhermore", I think most of this has been said, so maybe summarize with something like "An iterative method can thus be significantly cheaper than a direct method, by using a small number of iterations and taking advantage of matrix structure"

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+with a number of matrix-vector multiplications that is much lower than n or
+independent of n, it can outperform a direct method for the same problem.
+
+This means that iterative methods are most often useful in the following
```
btracey: s/This means that iterative/Iterative

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+situations:
+
+ - the system matrix A is sparse, blocked or has other special structure,
```
btracey: captalize bullets.

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+ - the problem size is sufficiently large that a dense factorization of A is too
```
btracey: don't need the word "too" I think

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+   costly in terms of computer time and/or storage,
+ - computing the product of A (or A^T, if necessary) with a vector can be done
```
btracey: maybe just "can be done efficiently"

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+   more efficiently than if A were first formed explicitly as a matrix,
+ - it suffices to find only an approximation to the solution.
```
btracey: An approximate solution is all that is required

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+
+Using linsolve
+
+The two most important elements of the API for solving linear systems are the
```
btracey: remove "for solving linear systems"

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+MulVecToer interface and the Iterative function.
+
+The MulVecToer interface represents the system matrix A in this package. It
+separates the method implementations from the details of any particular matrix
```
btracey: This abstracts the details of any particular matrix storage, and allows the user to exploit the properties of their particular matrix.

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+storage format. All matrix types provided by gonum.org/v1/gonum/mat and
```
btracey: s/by gonum/by the gonum

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+github.com/james-bowman/sparse packages implement this interface.
+
+Note that methods in this package have only limited means for checking whether
```
btracey: remove "whether"

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+the provided MulVecToer represents a matrix that satisfies all assumptions made
+by the chosen Method, for example whether the matrix is actually symmetric
```
btracey: for example if the matrix

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+ - computing the product of A (or A^T, if necessary) with a vector can be done
+   more efficiently than if A were first formed explicitly as a matrix,
+ - it suffices to find only an approximation to the solution.
+
+Using linsolve
+
+The two most important elements of the API for solving linear systems are the
+MulVecToer interface and the Iterative function.
+
+The MulVecToer interface represents the system matrix A in this package. It
```
btracey: Use a subheading. 

MulVecToer Interface:
The MulVecToer ...

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+separates the method implementations from the details of any particular matrix
+storage format. All matrix types provided by gonum.org/v1/gonum/mat and
+github.com/james-bowman/sparse packages implement this interface.
+
+Note that methods in this package have only limited means for checking whether
+the provided MulVecToer represents a matrix that satisfies all assumptions made
+by the chosen Method, for example whether the matrix is actually symmetric
+positive definite.
+
+The Iterative function is the entry point to the functionality provided by this
```
btracey: Iterative Function:

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+package. It takes as parameters the matrix A (as a MulVecToer), the right-hand
```
btracey: (via the MulVecToer interface as discussed above)

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+side vector b, the iterative method and settings that control the iterative
+process and provide a way for reusing memory.
+
+Choosing an iterative method
+
+The choice of an iterative method is typically guided by the properties of the
+matrix A including symmetry, definiteness, sparsity, conditioning and block
```
btracey: positive definiteness

btracey: conditioning, and block

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+structure. This is done to assure convergence of the algorithm for the given
```
btracey: This is done? What is done? This seems like a useless sentence, but maybe I don't know what you mean.

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+problem and efficiency of the computation. The methods of choice are well
```
btracey: Maybe something like "In general, performance on symmetric matrices is well understood (see the references below), with ConjugateGradient being a good starting point. Non-symmetric matrices are much more difficult to assess, where any suggestion of a 'best' method is usually accompanied by a recommendation to use trial-and-error.

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+understood for symmetric problems in which case their convergence rate (at least
+in exact arithmetics) is determined by spectral properties (eigenvalues) of A.
+The situation is not as clear for problems with non-symmetric matrices and it
+can be difficult to pick the best method for these problems. Any general advice
+for non-symmetric problem is often complemented with a trial-and-error approach.
+
+Preconditioning
+
+Preconditioning is a family of techniques that attempt to transform the linear
+system into one that has the same solution but more favorable spectrum. The
```
btracey: eigenspectrum?

linsolve/doc.go
```diff
@@ -3,28 +3,112 @@
...
+transformation matrix is called a preconditioner. A good preconditioner improves
```
btracey: A good preconditioner will reduce the number of iterations needed to find a good approximate solution (hopefully enough to overcome the cost of applying the preconditioning!), and in some cases preconditioning is necessary to get any kind of convergence.

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+// Setup initializes the receiver for solving the Allen-Cahn equation on a
+// uniform grid with n nodes on the spatial interval (0,L) and with the time
+// step size tau.
+func (ac *AllenCahnFD) Setup(n int, L float64, tau float64) {
+	ac.h = L / float64(n-1)
+	ac.tau = tau
+
+	// The finite difference scheme can be derived by replacing the derivatives
+	// by finite differences:
+	//  (u^k_i - u^{k-1}_i) / tau = (u^k_{i-1} - 2*u^k_i + u^k_{i+1}) / h² - 1/ξ² * f'(u^{k-1}_i)
```
btracey: t instead of k?

btracey: Maybe it's just me, but I always find implicit methods clearer when the next step is "t+1" and the values we currently have are "t"

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"gonum.org/v1/gonum/mat"
+	"gonum.org/v1/plot"
+	"gonum.org/v1/plot/plotter"
+	"gonum.org/v1/plot/plotutil"
+	"gonum.org/v1/plot/vg"
+)
+
+// AllenCahnFD implements a semi-implicit finite difference scheme for the
+// solution of the one-dimensional Allen-Cahn equation
+//  ∂_t u = Δu - 1/ξ²·f'(u)  in (0,L)×(0,T)
```
btracey: Could we replace the delta with del^2? I think the latter is much more common

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// Setup initializes the receiver for solving the Allen-Cahn equation on a
+// uniform grid with n nodes on the spatial interval (0,L) and with the time
+// step size tau.
+func (ac *AllenCahnFD) Setup(n int, L float64, tau float64) {
+	ac.h = L / float64(n-1)
+	ac.tau = tau
+
+	// The finite difference scheme can be derived by replacing the derivatives
```
btracey: Might be nicer to give a bit more detail, something like

"We can solve this PDE numerically by replacing the derivatives with finite differences. For the spatial derivative, we use a central difference scheme, and for the time derivative we use implicit euler integration.

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// by finite differences:
+	//  (u^k_i - u^{k-1}_i) / tau = (u^k_{i-1} - 2*u^k_i + u^k_{i+1}) / h² - 1/ξ² * f'(u^{k-1}_i)
+	// Collecting the terms from the same time level on each side of the
+	// equation gives:
+	//  u^k_i - tau / h² * (u^k_{i-1} - 2*u^k_i + u^k_{i+1}) = u^{k-1}_i - tau/ξ² * f'(u^{k-1}_i)
```
btracey: I think this would be clearer grouping as the u^k_{i-1}, u^k, u^k_{i+1} terms individually.

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// The zero Neumann boundary condition is imposed by reflecting the solution
```
btracey: Something like:

In an implicit scheme, the value for u^k_i (or in my notation u^{t+1}_i) depends on the value of u^{t+1}_{i-1} and u^{t+1}_{i+1}. Across the full spatial grid, the corresponding equations form a large linear system that must be solved to advance in time.

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// across the boundary:
+	//  u_{-1} = u_1,   u_{n+1} = u_{n-1}
+	// giving for example at i=0:
+	//  u^k_0 / 2 - tau / h² * (u^k_1 - u^k_0) = (u^{k-1}_0 - tau/ξ² * f'(u^{k-1}_0)) / 2
+
+	// Assemble the symmetric tridiagonal system matrix A based on the above
+	// discretization scheme.
+	const lda = 2
```
btracey: This might make sense for those of us who have spend way too much time in lapack, but maybe there's a better name?

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,200 @@
...
+	a := make([]float64, n*lda)
+	coef := tau / ac.h / ac.h
+	// Boundary condition at the left node
+	a[0] = 0.5 + coef
```
btracey: I think this would be much clearer if we allocated the SymBandDense and then used Set. It's not obvious (for instance), why it's two and not three.

linsolve/doc.go
vladimir-ch: I'd keep "finite" as a contrast to a potentially infinite number of steps of an iterative method to get to the exact solution.

linsolve/doc.go
vladimir-ch: I'm not sure what you mean. To write it as "The matrix, A, and the right-hand side, b, ..."?

linsolve/example_pde_test.go
vladimir-ch: I don't know what symbol you mean by "del", I think that "\Delta" is a common notation in PDEs for the Laplacian operator. There are also other common alternatives but since this is 1D, I changed the notation to the simple `u_t = u_xx - ...`.

linsolve/example_pde_test.go
vladimir-ch: I changed it to something similar.

linsolve/example_pde_test.go
vladimir-ch: I was aware of that and hoped to ask for an advice. I didn't want the test to produce output files but wanted to keep the plotting code. I've now commented out the `output` function and added a comment here but if you disagree I'm open to suggestions.

linsolve/doc.go
vladimir-ch: I think I'll have to remove this, it is not (yet?) true.

linsolve/example_pde_test.go
vladimir-ch: > t instead of k?

I kept `k` because I want to reserve `t` for the continuous time.

> the next step is "t+1" and the values we currently have are "t"

I had it with `k` and `k-1` because it reduced the amount of notation (there are many more terms on the next time level which now changed from `u^k` to `u^{k+1}`) but I changed it as you suggested.

linsolve/example_pde_test.go
vladimir-ch: Yes, good suggestion, thanks.

linsolve/doc.go
kortschak: Yes, but looking now, I think what you have is right.

linsolve/example_pde_test.go
kortschak: You can have a `if debug { output(ac.Solution(), L, i) }` where `debug' is a `const`. This won't get linked when it's false, and won't cause a linting complaint.

linsolve/example_pde_test.go
vladimir-ch: I now allocate and set `SymBandDense` directly without doing it the lapack style.

linsolve/example_pde_test.go
vladimir-ch: I've realized that if I remove `// Output:`, the example will not run but it will appear in godoc which is what I wanted.

vladimir-ch:
I think I've resolved all your comments, to some of them I responded explicitly. PTAL

linsolve/example_pde_test.go
kortschak: Yes. Without the `// Output:` comment, the example is built, but not run.

linsolve/bicg.go
```diff
@@ -10,10 +10,12 @@ import (
...
-// BiCG implements the Bi-Conjugate Gradient method with
-// preconditioning for solving systems of linear equations
-//  A*x = b,
-// where A is a square, possibly nonsymmetric matrix.
+// BiCG implements the Bi-Conjugate Gradient method with preconditioning for
+// solving systems of linear equations
+//  A * x = b,
+// where A is a nonsymmetric, nonsingular matrix. It uses limited memory storage
+// but the convergence may be irregular and the method may break down. BiCG
+// requires a multiplication with A and A^T at each iteration.
```
btracey: Maybe add a comment about BiCGStab?

linsolve/cg.go
```diff
@@ -8,10 +8,11 @@ import (
...
 
-// CG implements the Conjugate Gradient iterative method with
-// preconditioning for solving systems of linear equations
-//  A*x = b,
-// where A is a symmetric positive definite matrix.
+// CG implements the Conjugate Gradient iterative method with preconditioning
+// for solving systems of linear equations
+//  A * x = b,
+// where A is a symmetric positive definite matrix. It requires minimal memory
+// storage and is a good choice for symmetric positive definite problem.
```
btracey: s/problem/problems

linsolve/doc.go
```diff
@@ -3,28 +3,116 @@
...
+Iterative methods, in contrast, generally do not compute the exact solution x.
+Starting from an initial estimate x_0, they instead compute a sequence x_i of
+increasingly accurate approximations to x. This iterative process is stopped
+when the estimated difference between x_i and the true x becomes smaller than a
+prescribed threshold. The number of iterations thus depends on the value of the
+threshold. If the desired threshold is very small, then the iterative methods
+may take as long or longer than a direct method. However, for many problems a
+decent approximation can be found in a small number of steps.
+
+THe iterative methods implemented in this package do not access the elements of
```
btracey: s/THe/The

linsolve/doc.go
```diff
@@ -3,28 +3,116 @@
...
+cheaper depending on the structure of the matrix (sparse, banded,
+block-stuctured, etc.). Such structure often arises in practical applications.
+An iterative method can thus be significantly cheaper than a direct method, by
+using a small number of iterations and taking advantage of matrix structure.
+
+Iterative methods are most often useful in the following situations:
+
+ - The system matrix A is sparse, blocked or has other special structure,
+ - The problem size is sufficiently large that a dense factorization of A is
+   costly in terms of computer time and/or storage,
```
btracey: compute time and/or memory storage?

linsolve/doc.go
```diff
@@ -3,28 +3,116 @@
...
+ - An approximate solution is all that is required.
+
+Using linsolve
+
+The two most important elements of the API are the MulVecToer interface and the
+Iterative function.
+
+MulVecToer interface
+
+The MulVecToer interface represents the system matrix A in this package. This
```
btracey: I'm not sure we need "in this package"

linsolve/doc.go
```diff
@@ -3,28 +3,116 @@
...
+abstracts the details of any particular matrix storage, and allows the user to
+exploit the properties of their particular matrix. Matrix types provided by
+github.com/james-bowman/sparse packages implement this interface.
```
btracey: gonum/mat and github.com/james?

linsolve/example_iterative_test.go
```diff
@@ -0,0 +1,93 @@
+// Copyright ©2016 The Gonum Authors. All rights reserved.
```
btracey: 2019?

linsolve/example_iterative_test.go
```diff
@@ -0,0 +1,93 @@
...
+	grid := make([]float64, n+1)
+	for i := range grid {
+		grid[i] = x0 + float64(i)*h
+	}
+	grid[n] = x1
+	return grid
+}
+
+func ExampleIterative() {
+	grid := UniformGrid(0, 1, 10)
```
btracey: // Make a uniform grid
grid := make([]float64, n+1)
floats.Span(grid, x0, x1)

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,243 @@
...
+	//  u^k_i ≅ u(k*tau,i*h)
+	// Multiplying the equation by tau and collecting the terms from the same
+	// time level on each side gives
+	//  -tau/h²*u^{k+1}_{i-1} + (1+2*tau/h²)*u^{k+1}_i - tau/h²*u^{k+1}_{i+1}) = u^k_i - tau/ξ²*f'(u^k_i)
+	// If we denote C:=tau/h² we can simplify this to
+	//  -C*u^{k+1}_{i-1} + (1+2*C)*u^{k+1}_i - C*u^{k+1}_{i+1} = u^k_i - tau/ξ²*f'(u^k_i)   (1)
+	// which must hold for all i=0,...,n.
+	//
+	// When i=0 or i=n, the expression (1) refers to values at nodes that lie
+	// outside of the grid. We resolve this by using the zero Neumann boundary
```
btracey: maybe just "derivatives are zero" rather than zero Neumann? I also don't like the "we resolve this by", the condition comes from physics, but that kind of makes it seem like we made it up.

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,243 @@
...
+	//   u^{k+1}_{-1} = u^{k+1}_1
+	//  u^{k+1}_{n+1} = u^{k+1}_{n-1}
+	// If we substitute these two expressions into (1) at i=0 and i=n and divide
+	// by 2, we finally get
+	//  (1/2+C)*u^{k+1}_0 - C*u^{k+1}_1 = 1/2*u^k_0 - 1/2*tau/ξ²*f'(u^k_0)        (2)
+	//  -C*u^{k+1}_{n-1} + (1/2+C)*u^{k+1}_n = 1/2*u^k_n - 1/2*tau/ξ²*f'(u^k_n)   (3)
+	//
+	// For a given k the equations (1),(2),(3) form a linear system for the
+	// unknown vector [u^{k+1}_i], i=0,...,n that must be solved at each step in
+	// order to advance the solution in time. The matrix of this system is
```
btracey: This is fantastic. Could you also spell out b?

linsolve/example_pde_test.go
```diff
@@ -0,0 +1,243 @@
...
+	// condition - we approximate it by central difference:
+	//      -1/h*(u^{k+1}_{-1} - u^{k+1}_1) = 0
+	//  1/h*(u^{k+1}_{n+1} - u^{k+1}_{n-1}) = 0
+	// which after simplifying gives
+	//   u^{k+1}_{-1} = u^{k+1}_1
+	//  u^{k+1}_{n+1} = u^{k+1}_{n-1}
+	// If we substitute these two expressions into (1) at i=0 and i=n and divide
+	// by 2, we finally get
+	//  (1/2+C)*u^{k+1}_0 - C*u^{k+1}_1 = 1/2*u^k_0 - 1/2*tau/ξ²*f'(u^k_0)        (2)
+	//  -C*u^{k+1}_{n-1} + (1/2+C)*u^{k+1}_n = 1/2*u^k_n - 1/2*tau/ξ²*f'(u^k_n)   (3)
```
btracey: I would note something like "Note that this simply means we treat the edge values as the same as the inner values"

linsolve/bicg.go
vladimir-ch: Done.

linsolve/doc.go
vladimir-ch: I added it although it's not true yet and probably shouldn't be for all our types.

linsolve/example_iterative_test.go
vladimir-ch: I probably started writing it in 2016 but the first linsolve PR to exp was in 2017, so I changed it to that year.

linsolve/example_iterative_test.go
vladimir-ch: Of course, this is better. Thanks.

linsolve/example_pde_test.go
vladimir-ch: Good point, I hope it's better now.

linsolve/example_pde_test.go
vladimir-ch: Done.

linsolve/example_pde_test.go
vladimir-ch: Done.

PR #44, opened 2020-03-06
linsolve: use mat.VecDense instead of raw slices

vladimir-ch:
I think this turned out quite nicely, I'm happy especially with GMRES.

Please take a look.


linsolve/gmres.go
```diff
@@ -197,65 +187,85 @@ func (g *GMRES) Iterate(ctx *Context) (Operation, error) {
...
-		bi.Drot(1, hk[i:], 1, hk[i+1:], 1, givs[i].c, givs[i].s)
+// qr applies previous Givens rotations to the k-th column of H, computes the
+// next Givens rotation to zero out H[k+1,k] and applies it also to the vector s.
+func (g *GMRES) qr(k int, givs []givens, h *mat.Dense, s *mat.VecDense) {
+	// Apply previous Givens rotations to the k-th column of H.
+	hk := h.ColView(k).(*mat.VecDense)
+	for i, giv := range givs[:k] {
+		hki, hki1 := giv.apply(hk.AtVec(i), hk.AtVec(i+1))
+		hk.SetVec(i, hki)
+		hk.SetVec(i+1, hki1)
```
kortschak: Is there a reason for doing this in this package rather than using BLAS?

linsolve/gmres.go
```diff
@@ -197,65 +187,85 @@ func (g *GMRES) Iterate(ctx *Context) (Operation, error) {
...
+	// Apply the k-th Givens rotation to (H[k,k], H[k+1,k]).
+	hkk, _ := givs[k].apply(hk.AtVec(k), hk.AtVec(k+1))
+	hk.SetVec(k, hkk)
+	// We don't call hk.SetVec(k+1, 0) because the element won't be accessed.
+
 	// Apply the k-th Givens rotation to (s[k], s[k+1]).
-	bi.Drot(1, s[k:], 1, s[k+1:], 1, givs[k].c, givs[k].s)
+	sk, sk1 := givs[k].apply(s.AtVec(k), s.AtVec(k+1))
+	s.SetVec(k, sk)
+	s.SetVec(k+1, sk1)
```
kortschak: Same question.

linsolve/iterative.go
```diff
@@ -20,21 +20,21 @@ var ErrIterationLimit = errors.New("linsolve: iteration limit reached")
 // multiplication.
 type MulVecToer interface {
 	// MulVecTo computes A*x or Aᵀ*x and stores the result into dst.
-	MulVecTo(dst []float64, trans bool, x []float64)
+	MulVecTo(dst *mat.VecDense, trans bool, x mat.Vector)
 }
 
 // Settings holds settings for solving a linear system.
 type Settings struct {
 	// InitX holds the initial guess. If it is nil, the zero vector will be
```
kortschak: I would make this be "If it is empty, the zero vector will be". This way, pre-allocated but not size-conformed vectors can be used if `Reset` first.

linsolve/iterative.go
```diff
@@ -20,21 +20,21 @@ var ErrIterationLimit = errors.New("linsolve: iteration limit reached")
...
 	// used, otherwise its length must be equal to the dimension of the
 	// system.
-	InitX []float64
+	InitX *mat.VecDense
 
 	// Dst, if not nil, will be used for storing the approximate solution,
```
kortschak: Same here.

linsolve/iterative.go
```diff
@@ -57,18 +57,18 @@ type Settings struct {
...
-	// multiple linear systems. If Work is not nil, the length of its slice
-	// fields must be equal to the dimension of the system.
+	// multiple linear systems. If Work is not nil, its fields will be Reset
+	// prior to use.
 	Work *Context
 }
 
 // defaultSettings fills zero fields of s with default values.
 func defaultSettings(s *Settings, dim int) {
 	if s.Dst == nil {
```
kortschak:
```
if s.Dst == nil {
    mat.NewVecDense(dim, nil)
} else {
    s.Dst.ReuseVecAs(dim)
}
```

linsolve/bicg.go
```diff
@@ -24,36 +24,35 @@ import (
...
-	copy(b.x, x)
-	b.r = reuse(b.r, dim)
-	copy(b.r, residual)
-	b.rt = reuse(b.rt, dim)
-	copy(b.rt, b.r)
-	b.p = reuse(b.p, dim)
-	b.pt = reuse(b.pt, dim)
-	b.z = reuse(b.z, dim)
-	b.zt = reuse(b.zt, dim)
+	b.x.CloneVec(x)
```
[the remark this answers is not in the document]
vladimir-ch: Yes. Do we do it in exp or after merging to main?

linsolve/example_iterative_test.go
```diff
@@ -17,15 +17,7 @@ import (
 //  A*x = b
```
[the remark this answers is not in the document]
vladimir-ch: Ok, I wasn't aware. I'll rename it now.

linsolve/gmres.go
```diff
@@ -125,49 +113,47 @@ func (g *GMRES) Init(x, residual []float64) {
...
+		v0.ScaleVec(1/norm, v0)
 		// Initialize s to the elementary vector e_1 scaled by norm.
-		for i := range g.s {
-			g.s[i] = 0
-		}
-		g.s[0] = norm
+		g.s.Zero()
+		g.s.SetVec(0, norm)
 
 		// for k := 0; k < m; k++ {
```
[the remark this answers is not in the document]
vladimir-ch: Just an implementation comment trying to show where the for loop over k begins. I can remove it if you want.

linsolve/bicg.go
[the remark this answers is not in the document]
kortschak: In the merge to main, otherwise we need to update to a pseudo-version.

linsolve/gmres.go
```diff
@@ -197,65 +187,85 @@ func (g *GMRES) Iterate(ctx *Context) (Operation, error) {
...
+func (g *GMRES) modifiedGS(k int, h, v *mat.Dense, w *mat.VecDense) {
+	hk := h.ColView(k).(*mat.VecDense)
+	for j := 0; j <= k; j++ {
+		vj := v.ColView(j).(*mat.VecDense)
+		hkj := mat.Dot(vj, w)
+		hk.SetVec(j, hkj)           // H[j,k] = v_j · w
+		w.AddScaledVec(w, -hkj, vj) // w -= H[j,k] * v_j
 	}
-	wnorm := floats.Norm(w, 2)
-	hk[k] = wnorm            // H[k+1,k] = |v_{k+1}|
```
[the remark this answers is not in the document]
vladimir-ch: No, `modifiedGS` was receiving `g.k+1`, now it receives `g.k` which I think is clearer.

linsolve/gmres.go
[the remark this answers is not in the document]
kortschak: I'm happy to have that, but it should be in English; this looks as though it's a left over code fragment and so would be ignored.

linsolve/gmres.go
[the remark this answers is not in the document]
kortschak: Ah yes. That's way clearer now.

linsolve/gmres.go
vladimir-ch: Originally I was using `Drot` which is ok with raw slices (where increment is 1) but becomes onerous with `VecDense` - `Drot` operates on BLAS vectors, so I have to get the raw vector, then slice its `Data` twice with a general `Stride` and then pass the two **1-element** vectors to `Drot`. I think I prefer doing it locally as here.

linsolve/iterative.go

linsolve/iterative.go
vladimir-ch: Thanks, done.

vladimir-ch: Actually not yet done. `ReuseVecAs` panics for non-empty receiver, so unless we change that (to not panic if the vector already has the requested size), it has to be:
```
if s.Dst == nil {
    s.Dst = mat.NewVecDense(dim, nil)
} else if s.Dst.Len() != dim {
    s.Dst.ReuseAsVec(dim)
}
```

vladimir-ch: I've changed it as in my example above.

linsolve/gmres.go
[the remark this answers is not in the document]
vladimir-ch: Ok, done.

linsolve/iterative.go
vladimir-ch: Done but please take a look if the logic and the description are correct.

linsolve/iterative.go
kortschak: Thanks. Yes, that's how we have it elsewhere.

linsolve/gmres.go
kortschak: OK

linsolve/iterative.go
```diff
@@ -25,15 +25,14 @@ type MulVecToer interface {
 
 // Settings holds settings for solving a linear system.
 type Settings struct {
-	// InitX holds the initial guess. If it is nil, the zero vector will be
-	// used, otherwise its length must be equal to the dimension of the
-	// system.
+	// InitX holds the initial guess. If it is empty, the zero vector will be
```
kortschak: "If it is nil or empty..."

vladimir-ch: Done, plus a similar change for Dst and Work.

PR #48, opened 2021-02-26
Add  ARLS.go to Gonum

rejones7go:
Proposing to add code called ARLS.go.
The name means "Auto Regularized Linear Solver".
This package solves ill-conditioned systems of linear equations automatically.
The code is robust in that it has no algorithmic panics at all, as problematic linear systems are
considered to be its normal input.
See more about this package at http://www.rejones7.net/ArlsGo/index.htm .

Please take a look.

### DIFF
--- a/arls.go
+++ b/arls.go
@@ -1,0 +1,1199 @@
+package arls
+/*
+Copyright (c) 2020 The Go Authors. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+   * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+/*
+The routines in this package solve the linear system of equation, 
+    Ax = b
+for any shape matrix. (Some routines add constraint equations.)
+The system can be underdetermined, square, or over-determined.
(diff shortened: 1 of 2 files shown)


arls.go
```diff
@@ -0,0 +1,1199 @@
+package arls
+/*
+Copyright (c) 2020 The Go Authors. All rights reserved.
```
kortschak, line 3: Is it OK to change this to the Gonum license? If it is, please amend this to be 
```
// Copyright ©2021 The Gonum Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.
```

kortschak (commented):
Modulo the license query (if it needs to retain the Go license, that's fine fine though) I'm happy to merge this on the basis that we'll massage it into idiomatic Go. I'd certainly like you to be involved in reviewing the changes that we make to do that if that's OK with you, Ron.

Thanks for making this contribution.

rejones7go:
Sure. I didn’t realize gonum had a separate license.

kortschak:
Would you please make the license change?

kortschak (approved):
Thank you.

rejones7go:
Yes, will do.

Actually, I just changed the GO license to the brief 3-line GONUM license,
which references the LICENSE file.
I also did a "commit" for arls and test_arls.
Is that it? Is that all there is to do?
No new PR?

kortschak:
This is all good. Thanks for that. Soon I plan to go through the code and massage it into idiomatic Go. I'll cc you on changes.

arls.go
```diff
@@ -0,0 +1,1173 @@
...
+    //solve
+    sigma = 0.0
+    lambda = 0.0
+    check := 0.0
+    if ur >= nr {
+        x, check = rmslambda(b, &U, S, &V, ur, 0.0)
+    } else {
+        noise := Utb
+        noise.SliceVec(ur, mn)
+        sigma = myrms(noise)   //sigma := myrms(Utb[ur:mn])
```
kortschak, line 443: @rejones7go Can I get a clarification of this?

`SliceVec` does not slice the receiver, it returns a slice that shared the same backing data. So these three lines are equivalent to `sigma = myrms(Utb)`. I made a change to the code so that it reflects the comment, but that then fails.

```
--- FAIL: TestArlsall (0.00s)
	    arls_test.go:15: TestArlsAll(7) failed.
```

rejones7go:
That's a serious error on my part.
What I need is the root mean square of the elements from index ur to mn...
I will need to investigate that.
Please show me the line you changed it to.
Ron

kortschak:
To match the comment, the code should be
```
		noise := Utb.SliceVec(ur, mn).(*mat.VecDense)
		sigma = myrms(noise) // sigma := myrms(Utb[ur:mn])
```

rejones7go:
OK, thanks.
This fix is important.
The tests in test_arls() didn’t highlight the bug because all the tests are
using tiny systems.
I will commit a fix for test_arls (and arls) but it will take me a day or
two to make all the adjustments and add a “real world” problem.
Thanks a million for noticing this.
Ron

kortschak:
No worries. Note that you'll need to pull master since the code has been gofmt'd and moved into `arls/`.

rejones7go:
Dan... I just updated arls and test_arls.
Now, when I reintroduce the bug you found -- just to see how the tests
behave -- instead of NONE of the tests detecting, SIX of them do,

A couple of other things I want to mention:

1. I noticed that the copyright statement is not actually correct.
You can Google "Correct form of a copyright notice" to see the recommended
form.
Basically, it is redundant to say "Copyright © 2021 by ... "
The correct form is just "Copyright 2021..... " or else  "(c) 2021 by ...."
Doesn't matter to me, but you might want to think about it.

2. Since Arls has a fairly robust and efficient general inequality
constraint solver,
it only takes a dozen lines of code each to made routines which will :
-- force a solution to be non-decreasing (i.e., casually, "rising")
-- force a solution to be "falling"
-- force a solution to be concave up (or flat)
-- force a solution to be concave down (or flat)
I have done all these in my python code recently and they worked very well.
I wonder if you would think it would be worthwhile to add these little
extras?
Of course, they could all be wrapped in to one routine, depending on Go's
style for such things.

Thanks for everything!
Ron

kortschak:
Thanks, Ron.

Would you send a PR with those changes?

1. I doubt this is a significant concern since the intention is clear ('(c)' is really just a typographic workaround from the period of typesetting when '©' was not available just as underlining is a replacement for italics).

2. When @vladimir-ch has time to consider this, that would be an issue for him.

rejones7go:
Sorry... do you mean the ideas for rising/falling/concave etc?

rejones7go:
I guess you mean for the bug fix to arls and the resulting changes to
test_arls

Ron

kortschak:
Yeah, the changes for the bug fix.

PR #50, opened 2021-03-05
Fix ARLS slicevec bug and enhance tests

rejones7go:
Dan noticed a line of code that had no effect due to misuse of slicevec. 
The effect was very serious, causing excessive regularization in real world examples.
The line of code of repaired to calculate the noise level in the problem's right hand side correctly.

The problem was not noticed in test_arls because minimal size tests were used to elicit most possible bugs, and most answers were checked by examining only the magnitude of each problem residual (the final right-hand-side error) and not the solution itself. This change enhances many tests to include both the residual and the solution error, and tightens the tolerance.  Also, two larger examples were added that allow the regularization process to work more dramatically. If the bug were present now six of the tests would detect it.


kortschak (commented):
There are two copies of each file here, the copy of each in the root of the repository and the copy of each in arls. Can you reconcile this so that there are only the copies in arls?

rejones7go:
OK. I believe I got the old files deleted.

kortschak (approved):
Thanks

PR #53, opened 2021-03-18
Repair glitch in MyVecRandom

rejones7go:
1. In testing an extension to the current arls.go I noticed that the routine I was using for creating random errors was creating all positive errors. They should be balanced around zero. So i corrected MyVecRandom (also making it look to the eye a little more random,
2. This change caused three of the tests to need a larger tolerance, so these 3 were adjusted.
3. Some utility routines started with lower case characters, with no discernable reason. I made them al Upper Case.
Ron

Please take a look.

### DIFF
--- a/arls/arls_test.go
+++ b/arls/arls_test.go
@@ -1,6 +1,6 @@
 package arls
 
-// Copyright 2021 The Gonum Authors. All rights reserved.
+// Copyright @2021 The Gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 import (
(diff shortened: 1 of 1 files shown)


arls/arls_test.go
```diff
@@ -36,11 +36,11 @@ func MyMatPrint(A *mat.Dense) {
 	fmt.Println(" ")
 }
 
-func squareof(x float64) float64 {
+func Squareof(x float64) float64 {
```
kortschak, line 39: The uppercase initial indicates to the build chain that the function is available to external packages. In the final product, only the actual solvers will be exported.

kortschak (approved):

kortschak:
Sorry, this fell off my radar.

PR #54, opened 2021-05-21
add ivp package to solve multivariable, non-autonomous ODEs

soypat:
Previous discussions:

* https://github.com/gonum/gonum/issues/532
* https://groups.google.com/g/gonum-dev/c/gbsTBk2xY6M/m/bxZWbXJiBAAJ
* Recent PR of same package on gonum/gonum: https://github.com/gonum/gonum/pull/1641

I've made the relevant corrections. I'll make a proposal referencing this PR in the mailing list now. (see mailing list https://groups.google.com/g/gonum-dev/c/mBzLpX_MIgc)

### DIFF
--- a/ode/README.md
+++ b/ode/README.md
@@ -1,0 +1,36 @@
+# Gonum ivp [![GoDoc](https://godoc.org/gonum.org/v1/gonum/ivp?status.svg)](https://godoc.org/gonum.org/v1/gonum/ivp)
+
+package ode provides numerical methods solving multivariable ODE initial value problems for the Go programming language.
+
+Below is an example of usage
+```go
+	const (
+		g = -10. // gravity field [m.s^-2]
+	)
+	// we declare our physical model. First argument is initial time, which is 0 seconds.
+	// Next is the initial state vector, which corresponds to 100 meters above the ground
+	// with 0 m/s velocity.
+	ballModel, err := ode.NewIVP(0, mat.NewVecDense(2, []float64{100., 0.}),
+		func(yvec *mat.VecDense, _ float64, xvec mat.Vector) {
+			// this anonymous function defines the physics.
+			// The first variable xvec[0] corresponds to position
+			// second variable xvec[1] is velocity.
+			Dx := xvec.AtVec(1)
+			// yvec represents change in xvec, or derivative with respect to domain
+			// Change in position will be equal to velocity, which is the second variable:
+			// thus yvec[0] = xvec[1], which is the same as saying "change in xvec[0]" is equal to xvec[1]
+			yvec.SetVec(0, Dx)
+			// change in velocity is acceleration. We suppose our ball is on earth accelerating at `g`
+			yvec.SetVec(1, g)
+		})
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Here we choose our algorithm. Runge-Kutta 4th order is used
+	var solver ode.Integrator = ode.NewDormandPrince5(ode.DefaultParam)
+	// Solve function makes it easy to integrate a problem without having
+	// to implement the `for` loop. This example integrates the IVP with a step size
+	// of 0.1 over a domain of 10. arbitrary units, in this case, 10 seconds.
+	results, err := ode.SolveIVP(ballModel, solver, 0.1, 10.)
+	fmt.Println(results)
+```
--- a/ode/config.go
+++ b/ode/config.go
@@ -1,0 +1,11 @@
+package ode
+
+// Parameters represents the most common configuration parameters
+type Parameters struct {
+	// Permissible tolerance given an adaptive method
+	AbsTolerance float64
+	// Minimum/Maximum step allowed for a single iteration
+	MinStep, MaxStep float64
+}
+
+var DefaultParam = Parameters{}
--- a/ode/doc.go
+++ b/ode/doc.go
@@ -1,0 +1,7 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// package ode provides interfaces and structures to solve
+// multivariable ODE initial value problems.
+package ode // import "gonum.org/v1/gonum/ode"
--- a/ode/dopri5.go
+++ b/ode/dopri5.go
@@ -1,0 +1,172 @@
+package ode
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+type DoPri5 struct {
+	x, aux *mat.VecDense
+	// solutions
+	y4, y5, err45 *mat.VecDense
+	// integration coefficients (for Butcher Tableau)
+	k1, k2, k3, k4, k5, k6, k7 *mat.VecDense
+	dom                        float64
+	fx                         func(y *mat.VecDense, t float64, x mat.Vector)
+
+	// adaptive control parameters
+	atol, minStep, maxStep float64
+	adaptive               bool
+}
+
+// Set implements the Integrator interface. Initializes a Dormand Prince method
+// for use as a solver
+func (dp *DoPri5) Init(ivp IVP) {
+
+	dp.fx = ivp.Func
+
+	t0, x0 := ivp.T0, mat.VecDenseCopyOf(ivp.Y0)
+	dp.dom = t0
+	nx := x0.Len()
+	dp.k1, dp.k2, dp.k3 = mat.NewVecDense(nx, nil), mat.NewVecDense(nx, nil), mat.NewVecDense(nx, nil)
+	dp.k4, dp.k5, dp.k6 = mat.NewVecDense(nx, nil), mat.NewVecDense(nx, nil), mat.NewVecDense(nx, nil)
+	dp.k7 = mat.NewVecDense(nx, nil)
+	dp.aux = mat.NewVecDense(nx, nil)
+	dp.x = mat.NewVecDense(nx, nil)
+	dp.x.CloneFromVec(x0)
+
+	dp.y4 = mat.NewVecDense(nx, nil)
+	dp.y5 = mat.NewVecDense(nx, nil)
+	dp.err45 = mat.NewVecDense(nx, nil)
+
+}
+
+func (dp *DoPri5) State(s *State) {
+	s.T = dp.dom
+	s.Y.CloneFromVec(dp.x)
+}
+
+// Step implements Integrator interface. Advances solution by step h. If algorithm
+// is set to adaptive then h is just a suggestion
+func (dp *DoPri5) Step(h float64) (float64, error) {
+	const c20, c21 = 1. / 5., 1. / 5.
+	const c30, c31, c32 = 3. / 10., 3. / 40., 9. / 40.
+	const c40, c41, c42, c43 = 4. / 5., 44. / 45., -56. / 15., 32. / 9.
+	const c50, c51, c52, c53, c54 = 8. / 9., 19372. / 6561., -25360. / 2187., 64448. / 6561., -212. / 729.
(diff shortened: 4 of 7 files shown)


ivp/ivp.go
```diff
@@ -0,0 +1,108 @@
...
+//
+// These problems have the form (capital letters are vectors in this example)
+//  X'(t) = F(t, X, U)
+//  X(0) = X_0
+//
+// Where X' is the vector of first derivatives of the state vector X. t is
+// a scalar representing the integrations domain, which is usually time
+// for most physical problems. U is a vector which is a function of the current
+// state. Put simply:
+//  U = F_u(t, X, U)
```
vladimir-ch, line 27: What is this statement? I assume it's not an equation but if it's a definition of U, then 1) U should not appear on the right hand side and 2) what is F_u? Is the _u subscript an indication it's a Jacobian of F w.r.t. the third argument or is it just a vector function different from F? If U is a function of the current state (i.e., U(t) := G(X(t))), why can't it be folded into F so that the system becomes the usual X'(t)=F(t, X(t))? Are you following any reference, for example a book?

soypat: This is a common way of writing physical systems for their control. I believe these are [non-autonomous systems](https://en.wikipedia.org/wiki/Autonomous_system_(mathematics)). See _Feedback Control of Dynamic Systems_ by Gene Franklin or _Modern control engineering_ by Katsuhiko Ogata. The advantage of such description is that it permits separating the physical system's natural response from external factors which are actuated by the controller or external stochastic perturbations which cannot be described by a differential equation.

The advantage of such a description is that the System matrix (Jacobian of `F`) and the control matrix (Jacobian of `F_u`) are readily calculated from the returned parameters of `Equations()`.

soypat:
@vladimir-ch I've added a draft of Dormand Prince solver. Let me know what you think!

ivp/ivp.go
soypat: I don't have an exact page number for both of those books. If you want a condensed version I've taken some notes on the subject and compiled them here:
https://github.com/soypat/whittileaks/blob/master/control/control_rapido.pdf
see section 1.1.2 (page 4)

vladimir-ch: We want to create a general IVP solver without a preference for any potential applications, for example in control theory, or without assuming that there is even an underlying physical system. Things like "response", "actuated", "controller", "control matrix" are out of our scope.
 
Chapter 2 of Ogata is available as a sample. If you look at page 31:
```
x'(t) = f(x,u,t)           (2-10)
 y(t) = g(x,u,t)           (2-11)
```
where u is the input, in other words a known function of t. (2-10) is a differential equation for the unknown vector function x (the state). _Our task is to create a solver (or solvers) for finding x_. Once we solve it, (2-11) becomes just a recipe how to get y from x, u and t. So (2-11) is not relevant to our task and we can ignore it, only (2-10) is relevant. In (2-10), given a t, u(t) is known and so it is just a parameter of f and, again, not relevant to our task. As far as we are concerned, u is part of the definition of f.

And so we have reduced (2-10), (2-11) to only what is relevant to our task:
```
x'(t) = f(x, t)
```
(plus the initial condition). u and y are objects related to application-specific (control systems in this case) modeling. If you're interested in control theory, then later there can be another dedicated package built on top of the ivp (or whatever the name will be) package with concepts like "the input" and "the output".

soypat: Alright, I can respect the position of wanting to create a general IVP solver with no preference. You are right for the most part, getting `u` is a recipe which is a function of `t` *and* `x`.

soypat:
@vladimir-ch `u`, `ufunc` eliminated from package. Best to keep the base implementation simple like you said!

soypat:
I read into the `dopri5.f` code and liked the configurability. I've added a `Configuration` object to be able to modify solver behaviour. My motivation was to allow configurations to be portable across different solvers.

vladimir-ch:
Please don't add more code to this PR, I haven't even reviewed what was originally in. My expectation was that we would discuss, explore and iterate on the package design to find out what works and what doesn't and that can't be done if the goal constantly moves. I want to limit the functionality to explicit R-K solvers (a minimal viable product) because implicit solvers open a whole new world of complexity related to linear solvers and how to represent linear systems well in the package (in addition to a robust implementation, of course).

ivp/ivp.go
```diff
@@ -12,47 +12,29 @@ import (
...
+// for most physical problems.
 //
 // The term "state vector" and "state variables" are used interchangeably
 // throughout the code and refer to X vector of independent variables.
 type IVP interface {
-	// Initial values vector for state variables x and inputs u. x0 defines
+	// Initial values vector for state variables x. x0 defines
 	// the first values the state vector takes when integration begins.
-	IV() (x0, u0 mat.Vector)
+	IV() (x0 mat.Vector)
```
soypat, line 35: Maybe should return domain point at which initial values are defined. This would also change `Integrator`'s Step() signature.

ivp/ivp.go
```diff
@@ -0,0 +1,96 @@
...
+	// Set initializes an initial value problem. First argument
+	// is the initial domain integration point, is usually zero.
+	Set(float64, IVP) error
+	// Length of state vector x
+	XLen() (nx int)
+	// Step integrates IVP and stores result in y. step is a suggested step
+	// for the algorithm to take. The algorithm may decide that it is not sufficiently
+	// small or big enough (these are adaptive algorithms) and take a different step.
+	// The resulting step is returned as the first parameter
+	Step(y []float64, step float64) (float64, error)
```
vladimir-ch, line 60: I'd say this is the key line in this PR. It implies that the Integrator does the following things:

1. It changes the internal state of Integrator while trying to advance the solution. This means:
- apply the step size limitation (maybe, see below)
- compute new candidate solution
- in case of adaptive methods, estimate the error, redo and adapt the step if rejected, compute new step size if accepted
2. It returns the reached solution in `y`. Here we see a limitation of this interface: the caller doesn't know to what `x` `y` corresponds. It means that he must track the current `x` using the returned step size but since Integrator surely must track `x` internally as well, it looks like unnecessary extra work with a risk of differing values for current `x`. Maybe there should be a `State` type holding `x` and the corresponding `y` together.
3. It returns the used step size. This is a useful information but it leaves a bit unclear who is responsible for limiting the step size. Is it Integrator or its user?

I would use `mat.VecDense`, therefore:
```
Step(dst *mat.VecDense, step float64) (float64, error)
```

ivp/ivp.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type result = struct {
```
vladimir-ch, line 63: This could be the `State` type:
```
type State struct {
    X float64
    Y mat.VecDense
}
```

ivp/ivp.go
```diff
@@ -0,0 +1,96 @@
...
+	DomainStartOffset float64
+	X                 []float64
+}
+
+// Solve solves an already initialized Integrator returning state vector results.
+// Returns error upon needing to allocate over 2GB of memory
+func Solve(solver Integrator, stepsize, domainLength float64) (results []result, err error) {
```
vladimir-ch, line 70: With my proposals, this could be
```
func SolveIVP(p IVP, to float64, solver Integrator) (result []State, err error)
```
but we should think how the user will use this package.

This function means that the users knows the end x which does not need to be the case. Maybe he wants to integrate only until something happens (like a component crossing a value, zero).

It collects and returns _all_ states which can be many but very likely the user doesn't need states at basically arbitrary location determined by the step adaptivity. Maybe the user needs the solution at some predetermined locations between the steps. This can be achieved by repeated calls to this function but is this the best API (apart from other subtle questions)?

ivp/ivp.go
```diff
@@ -0,0 +1,96 @@
...
+	const maxAllocGB = 2
+	integrated := 0.
+	expectedLength := int(domainLength/stepsize) + 1
+	nx := solver.XLen()
+	if nx == 0 {
+		return nil, errors.New("state vector length can not be equal to zero. Has ivp been set?")
+	}
+	size := 8 * (nx + 1) * expectedLength
+	if size > maxAllocGB*1e9 {
```
vladimir-ch, line 79: Remove this.

ivp/ivp.go
```diff
@@ -0,0 +1,96 @@
...
+	for integrated < domainLength {
+		res := make([]float64, nx)
+		stepsize, err = solver.Step(res, stepsize)
+		if err != nil {
+			return results, err
+		}
+		if stepsize <= 0 {
+			return results, errors.New("got zero or negative step size from Integrator")
+		}
+		integrated += stepsize
```
vladimir-ch, line 92: On the topic of step limiting: this code suggests that it's Integrator's task to limit it but then it must know about the end x otherwise this code will integrate past `to` which is unexpected. As I wrote, writing an ODE solver is not easy and there are many subtle details.

ivp/model.go
```diff
@@ -0,0 +1,46 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ivp
+
+import "gonum.org/v1/gonum/mat"
+
+// Model implements IVP interface. X vector and equations can not be nil or zero length.
+type Model struct {
```
vladimir-ch, line 10: With `IVP` being a struct there is no need for this type/file.

ivp/newton.go
```diff
@@ -0,0 +1,115 @@
+package ivp
```
vladimir-ch, line 1: Please remove this code and its test from this PR, we're not ready to discuss implicit methods.

ivp/ivp.go
vladimir-ch: Yes, see my review comments.

ivp/doc.go
```diff
@@ -0,0 +1,7 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package ivp provides interfaces and structures to solve
+// multivariable ODE initial value problems.
+package ivp // import "gonum.org/v1/gonum/ivp"
```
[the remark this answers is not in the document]
soypat: Agreed, ode does it better

ivp/ivp.go
```diff
@@ -0,0 +1,96 @@
...
+// X' is the vector of first derivatives of the state vector X.
+// F would be xequations as returned by Equations().
+// An initial value problem is characterized by the initial conditions imposed
+// on the state vector X at the beginning of the integration domain. These
+// initial conditions are returned by the IV() method for the state vector
+// as x0 and for the input vector as u0.
+//
+// The term "state vector" and "state variables" are used interchangeably
+// throughout the code and refer to X vector of independent variables.
+type IVP interface {
```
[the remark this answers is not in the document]
soypat: I like that this simplifies a lot of things on the gonum/ode end. IVP could be embedded into Integrators and simplify some setup.

If we are sure we are not losing much by eliminating the interface abstraction, it is fine by me.

soypat:
I've [applied changes (d2407cb)](https://github.com/gonum/exp/pull/54/commits/d2407cbaf9235ff9f3c566e5c5ae94a3f59195ae) you proposed to drive the package closer to the goal. Some changes were not integrated because they still seem to need community discussion.

vladimir-ch:
I have to put off my review of this PR. I'm starting to have again too many pending tasks on my side and this PR and ode design discussion is not something that can be finished quickly. I need time to implement a couple of solvers myself to understand how the API should look like, what kind of information needs to be passed between Integrator and the driving routine, which side should do the step control, how to report errors, if and how to fit fixed-step solvers into it, implement a solid testing suite, etc.

soypat:
Please take a look. I have squashed and applied changes suggested.

vladimir-ch:
Thanks, I'll be able to take a look in ~2 weeks.

PR #57, opened 2022-03-29
ode: add Runge-Kutta-Nyström 12(10) method

soypat:
@vladimir-ch  Please take a look. New integrator and fixes to documentation.

This PR implements a method using a different integration scheme from the Integrator interface. 
```
// The existing Integrator interface solves the following problem:
//  y'(t) = f(t, y(t))
//  y(t_0) = y_0

// The proposed Runge-Kutta-Nyström integrator solves a problem of the form
//  y''(t)   = f(t, y(t))
//  y(t_0)   = y_0
//  y'(t_0)  = y'_0
```
The intention is to open a discussion on other kinds of integrators for different kind of ODEs.

**The algorithm has not been tested yet.**


ode/rkn1210.go
```diff
@@ -0,0 +1,117 @@
...
+	rk.fx = ivp.Func
+	rk.dom, rk.y = ivp.T0, mat.VecDenseCopyOf(ivp.Y0)
+	rk.dy = mat.VecDenseCopyOf(ivp.DY0)
+	ny := rk.y.Len()
+	rk.aux = mat.NewVecDense(ny, nil)
+	rk.hFbhat = mat.NewVecDense(ny, nil)
+	rk.hFb = mat.NewVecDense(ny, nil)
+	for j := 0; j < rk1210Len; j++ {
+		rk.f[j] = mat.NewVecDense(ny, nil)
+	}
```
sbinet, line 28:
```suggestion
	for j := range rk.f {
		rk.f[j] = mat.NewVecDense(ny, nil)
	}
```

ode/rkn1210.go
```diff
@@ -0,0 +1,117 @@
...
+	rk := new(RKN1210)
+	if cfg.AbsTolerance == 0 {
+		cfg.AbsTolerance = 1e-3
+	}
+	if cfg.MinStep == 0 {
+		cfg.MinStep = 1e-4
+	}
+	if cfg.MaxStep == 0 {
+		cfg.MaxStep = 1e3
+	}
```
sbinet, line 114: this doesn't seem to be very useful, as neither `cfg.MinStep` nor `cfg.MaxStep` are used later on in this function.

sbinet (commented):
just a couple of drive-by comments, while I was looking at it (out of curiosity, I am by no means an expert)

ode/rkn1210.go
soypat: Ok, I added a test and adaptive stepping to it so now they should be useful!

soypat:
Thanks @sbinet for taking the time to look at the PR! I have added a test and it seems to be working with 1dof problem.

soypat:
I can confirm, RKN1210 works **incredibly well**. I'm working with steps 5 orders of magnitude larger than with another RK method I was using beforehand.

soypat:
@kortschak This PR is ready for review! Vladimir seems to be AFK so I'm reaching out to you.

ode/ivp.go
```diff
@@ -47,3 +47,27 @@ func NewIVP(t0 float64, y0 mat.Vector, f func(y *mat.VecDense, dom float64, x ma
...
+type IVP2 struct {
+	Y0  mat.Vector
+	DY0 mat.Vector
+	T0  float64
+	// Func are the second derivatives of the solution such that
+	//  dst = y''(t) = Func(t, y(t))
+	Func func(dst *mat.VecDense, t float64, y mat.Vector)
+}
+
+// NewIVP2 returns a second order initial value problem
```
kortschak, line 67: Nit: final dot.

Also, given that the user can construct the value themselves without the use of this function it might be worth noting that it validates the parameters (that presumably being the value that it adds).

ode/ivp_test.go
```diff
@@ -141,6 +141,50 @@ func exp1DTestModel(t *testing.T) *TestModel {
 	return Quadratic
 }
 
+func TestRK1210_fallingball(t *testing.T) {
+	const (
+		tol     = 1e-10
+		dt      = .5
+		gravity = -10.
```
kortschak, line 148: 0.5 and -10.0 (or -9.8 since that's what the comment says).

ode/ivp_test.go
```diff
@@ -141,6 +141,50 @@ func exp1DTestModel(t *testing.T) *TestModel {
...
+		initPos = 10
+		initVel = 0.0
+	)
+	// Falling ball from initPos height at initial velocity initVel.
+	ivp := ode.IVP2{
+		Y0:  mat.NewVecDense(1, []float64{initPos}),
+		DY0: mat.NewVecDense(1, []float64{initVel}),
+		Func: func(dst *mat.VecDense, t float64, y mat.Vector) {
+			dst.SetVec(0, gravity) // gravity is -9.8m/s
```
kortschak, line 157: The comment is not necessary given the constant.

ode/ode.go
```diff
@@ -30,6 +30,12 @@ type State struct {
 	Y *mat.VecDense
 }
 
+type State2 struct {
```
kortschak, line 33: This needs a comment.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
+package ode
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+const rk1210Len = 17
```
kortschak, line 9: This needs a comment. Also, please place it below its use (this makes the comment less necessary, but it would still be good to have a provenance for its origin).

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
+package ode
```
kortschak, line 1: License header.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+
+const rk1210Len = 17
+
+// RKN1210 Runge-Kutta-Nyström integrator of order 12(10).
+type RKN1210 struct {
+	dom        float64
+	y, dy, aux *mat.VecDense
+	// Low and high order terms from integration.
+	hFDb, hFb, hFDbhat, hFbhat *mat.VecDense
+	f                          [rk1210Len]*mat.VecDense
```
kortschak, line 17: I think this could have a better name.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+	// Step control.
+	atol, minStep, maxStep float64
+	fx                     func(y *mat.VecDense, t float64, x mat.Vector)
+}
+
+func (rk *RKN1210) Init(ivp IVP2) {
```
kortschak, line 23: Comment.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+	rk.hFDbhat = mat.NewVecDense(ny, nil)
+	rk.hFbhat = mat.NewVecDense(ny, nil)
+	rk.hFDb = mat.NewVecDense(ny, nil)
+	rk.hFb = mat.NewVecDense(ny, nil)
+	for j := range rk.f {
+		rk.f[j] = mat.NewVecDense(ny, nil)
+	}
+}
+
+func (rk *RKN1210) State(s *State2) {
```
kortschak, line 38: Comment.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+	hFbH := rk.hFbhat
+	// Low order B's for error estimation
+	hFDb := rk.hFDb
+	hFb := rk.hFb
+
+	fun := rk.fx
+	F := rk.f
+	t := rk.dom
+
+SOLVE:
```
kortschak, line 92: Lower case.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+func (rk *RKN1210) State(s *State2) {
+	s.T = rk.dom
+	s.Y.CloneFromVec(rk.y)
+	s.DY.CloneFromVec(rk.dy)
+}
+
+// Step implements Integrator interface. Advances solution by step h. If algorithm
+// is set to adaptive then h is just a suggestion
+func (rk *RKN1210) Step(h float64) (float64, error) {
+	var (
```
kortschak, line 47: These can be put in the package scope.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+	)
+	adaptive := rk.atol > 0
+	// Declare shorthand variable names.
+	y := rk.y
+	dy := rk.dy
+	aux := rk.aux
+	// high order B's
+	hFDbH := rk.hFDbhat
+	hFbH := rk.hFbhat
+	// Low order B's for error estimation
```
kortschak, line 84: Final dot.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+	hFb.Zero()
+	hFDb.Zero()
+	h2 := h * h
+	for j := range F {
+		// aux = y + h*c[j] + F*h*h*a[j]
+		hc := h * c[j]
+		aux.AddScaledVec(y, hc, dy)
+		for iF := 0; iF < j; iF++ {
+			aux.AddScaledVec(aux, h2*A[j][iF], F[iF])
+		}
```
kortschak, line 104:
```
		for i, iF := range F[:j] {
			aux.AddScaledVec(aux, h2*A[j][i], iF)
		}
```

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+	F := rk.f
+	t := rk.dom
+
+SOLVE:
+	hFDbH.Zero()
+	hFbH.Zero()
+	hFb.Zero()
+	hFDb.Zero()
+	h2 := h * h
+	for j := range F {
```
kortschak, line 98:
```
	for j, jF := range F {
```

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+		// aux = y + h*c[j] + F*h*h*a[j]
+		hc := h * c[j]
+		aux.AddScaledVec(y, hc, dy)
+		for iF := 0; iF < j; iF++ {
+			aux.AddScaledVec(aux, h2*A[j][iF], F[iF])
+		}
+		// finally F[:,j] = Func( aux ) @ t+h*c[j]
+		fun(F[j], t+hc, aux) // 17 function evaluations.
```
kortschak, line 106:
```
		fun(Fj, t+hc, aux) // 17 function evaluations.
```
and similar below.

ode/rkn1210.go
```diff
@@ -0,0 +1,160 @@
...
+		if errMax > rk.atol && h > rk.minStep {
+			// Error is not permissible and we may redo the step.
+			h = hnew
+			goto SOLVE
+		}
+		// The error is within tolerance and we may suggest the user use a larger step.
+		// Modify return value to suggest new step.
+		defer func() { h = hnew }()
+	}
+	// calculate next step solutions with high order B's:
```
kortschak, line 139: Calculate

kortschak (commented):
This is just a superficial review at this stage.

soypat:
Thanks for the review @kortschak- that should do the trick!

ode/config.go
```diff
@@ -1,3 +1,7 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
```
kortschak, line 1: 2022?

ode/dopri5.go
```diff
@@ -1,3 +1,7 @@
+// Copyright ©2021 The Gonum Authors. All rights reserved.
```
kortschak, line 1: Same here.

ode/ivp_test.go
```diff
@@ -32,7 +32,7 @@ func TestSolve(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// run the solver and corsscheck values
+	// run the solver and crosscheck values.
```
kortschak, line 35: uc/r/

ode/ivp_test.go
```diff
@@ -56,7 +56,7 @@ func TestSolve(t *testing.T) {
 
 func Example_solve() {
 	const (
-		g = -10. // gravity field [m.s^-2]
+		g = -10.0 // local gravity field [m.s^-2]
```
kortschak, line 59: Is there a reason not to use the more accurate 9.8?

ode/rkn1210.go
```diff
@@ -0,0 +1,185 @@
...
+	// Low and high order terms from integration.
+	hFDb, hFb, hFDbhat, hFbhat *mat.VecDense
+	// second order differential equations.
+	diffs [rk1210Len]*mat.VecDense
+	// Step control.
+	atol, minStep, maxStep float64
+	fx                     func(y *mat.VecDense, t float64, x mat.Vector)
+}
+
+// Amount of Runge-Kutta iteration steps to calculate both high and low order solutions.
```
kortschak, line 26: s/Amount/Number/

ode/rkn1210.go
```diff
@@ -0,0 +1,185 @@
...
+
+// Step implements Integrator interface. Advances solution by step h. If algorithm
+// is set to adaptive then h is just a suggestion.
+func (rk *RKN1210) Step(h float64) (float64, error) {
+	A := &rkn12A
+	c := &rkn12c
+	b := &rkn12b
+	bp := &rkn12bp
+	bhat := &rkn12bhat
+	bphat := &rkn12bphat
```
kortschak, line 69: Why do these? They don't seem necessary.

ode/rkn1210.go
```diff
@@ -0,0 +1,185 @@
...
+	y.AddScaledVec(y, h, aux)
+	dy.AddVec(dy, hFDbH)
+	rk.dom += h
+	return h, nil
+}
+
+// NewRKN1210 configures a new RKN1210 instance.
+func NewRKN1210(cfg Parameters) *RKN1210 {
+	if (cfg.AbsTolerance != 0 && cfg.MaxStep <= 0) || cfg.MaxStep < cfg.MinStep ||
+		cfg.MinStep < 0 {
```
kortschak, line 147: One line.

ode/config.go
soypat: Even though the file was written in 2021? Shall I set it to 2022?

ode/ivp_test.go
soypat: 10m/s2 yields "cleaner" results of the balls trajectory that makes it easy for a curious onlooker to deduce it works as expected. But yeah, I could add this is the local gravity field of a hypothetical planet

ode/rkn1210.go
soypat: Without these some expressions below would be harder to read. I personally prefer having these short-hand declarations up here and 
```go
// this
hFDbH.AddScaledVec(hFDbH, h*bphat[j], Fj)
// instead of this:
hFDbH.AddScaledVec(hFDbH, h*rkn12bphat[j], Fj)
```
That said, I'm not married to it. We can use the latter form if you'd prefer.

kortschak: It just seems that an indirection is unnecessary.

ode/config.go
kortschak: Your copyright goes back to when you wrote it, Gonum's starts when it was contributed which starts about a month ago.

soypat:
All changes applied.

kortschak (dismissed):
Approving at exp level only since I'm not over this area.

ode/rkn1210.go
```diff
@@ -0,0 +1,177 @@
...
+	// high order B's.
+	hFDbH := rk.hFDbhat
+	hFbH := rk.hFbhat
+	// Low order B's for error estimation.
+	hFDb := rk.hFDb
+	hFb := rk.hFb
+
+	fun := rk.fx
+	F := rk.diffs
+	t := rk.dom
```
kortschak, line 78: I'm still not convinced that this is necessary.

ode/rkn1210.go
```diff
@@ -0,0 +1,177 @@
...
+	rk.dy = mat.VecDenseCopyOf(ivp.DY0)
+	ny := rk.y.Len()
+	rk.aux = mat.NewVecDense(ny, nil)
+	rk.hFDbhat = mat.NewVecDense(ny, nil)
+	rk.hFbhat = mat.NewVecDense(ny, nil)
+	rk.hFDb = mat.NewVecDense(ny, nil)
+	rk.hFb = mat.NewVecDense(ny, nil)
+	for j := range rk.diffs {
+		rk.diffs[j] = mat.NewVecDense(ny, nil)
+	}
```
kortschak, line 43: I think this may be better done with a `mat.Dense`; it will have a single slab allocation.

kortschak (commented):
Sorry to come back to this, I have found some time to take a closer look.

Can you add a reference for the method?

ode/rkn1210.go
soypat: Agreed although am unable to contribute this.

ode/rkn1210.go
soypat: Code does not hurt as much as I thought it would without those declarations. It is done :)

ode/rkn1210.go
kortschak: No worries. I'll take a look at how this would work.

ode/rkn1210.go
kortschak: Thanks

kortschak (approved):

PR #58, opened 2022-05-01
ode: use block allocation and fix step return

kortschak:
This makes a single slab allocation for the second order differential equation vectors and fixes the suggested step return (previously the deferred assignment was effectual since the return was not named), also improving the performance of this by not using defer.

@soypat There is no testing for this case (adaptive), so I don't know whether this is working - it should be if the previous version was.

Please take a look.


soypat:
LGTM

PR #61, opened 2023-06-03
fixes for SolveIVP implementation

soypat:
Please take a look. Fixes issue raised in mail group: https://groups.google.com/g/gonum-dev/c/_B0R3518KNI

also adds initial conditions as part of result vector returned by SolveIVP as that seems reasonably intuitive.


kortschak:
What is the fix?

ode/ode.go
```diff
@@ -46,10 +48,12 @@ func SolveIVP(p IVP, solver Integrator, stepsize, tend float64) (results []State
 	if nx == 0 {
 		return nil, errors.New("state vector length can not be equal to zero. Has ivp been set?")
 	}
+	solver.Init(p)
```
soypat, line 51: @kortschak Forgot to initialize the integrator. This was the pressing problem that caused the crash since it meant some fields were nil upon attempting to integrate.

kortschak: Thanks, that was my suspcion. Can you make a note of that in the commit message?

kortschak (approved):

ode/ode.go
soypat: alright!

PR #65, opened 2025-03-05
root: add Brent's method for function root finding

fumin:
This change also introduces a new package: root.
The root package is intended to contain algorithms for function root finding.
Aside from Brent's method, another candidate algorithm to be included in the root package is Newton's method.

Fixes https://github.com/gonum/gonum/issues/483 partially.

Please take a look.

### DIFF
--- a/root/brent.go
+++ b/root/brent.go
@@ -1,0 +1,105 @@
+package root
+
+import (
+	"errors"
+	"math"
+)
+
+// Brent finds the root of a function using Brent's method.
+// The root to be found should lie between [a, b], and will be refined until its accuracy is tol.
+// This implementation is based on page 352, Section 9.3, Vol 1: Numerical Recipes in Fortran 77 2nd Ed., William H. Press, Saul A. Teukolsky, William T. Vetterling, Brian P. Flannery.
+// https://s3.amazonaws.com/nrbook.com/book_F210.html
+//
+// See https://en.wikipedia.org/wiki/Brent%27s_method for more details.
+func Brent(f func(float64) float64, a, b, tol float64) (float64, error) {
+	const itmax = 100
+	var eps = math.Nextafter(1, 2) - 1
+
+	var d, e, xm float64
+
+	fa, fb := f(a), f(b)
+	if (fa > 0 && fb > 0) || (fa < 0 && fb < 0) {
+		return math.NaN(), errors.New("root must be bracketed")
+	}
+
+	c, fc := b, fb
+	for iter := 0; iter < itmax; iter++ {
+		if (fb > 0 && fc > 0) || (fb < 0 && fc < 0) {
+			// Rename a, b, c and adjust bounding interval d.
+			c, fc = a, fa
+			d = b - a
+			e = d
+		}
+		if math.Abs(fc) < math.Abs(fb) {
+			a, fa = b, fb
+			b, fb = c, fc
+			c, fc = a, fa
+		}
+
+		// Convergence check.
+		tol1 := 2*eps*math.Abs(b) + 0.5*tol
+		xm = 0.5 * (c - b)
+		if math.Abs(xm) < tol1 || fb == 0 {
+			return b, nil
+		}
+
+		if math.Abs(e) >= tol1 && math.Abs(fa) > math.Abs(fb) {
+			// Attempt inverse quadratic interpolation.
+			var p, q float64
+			s := fb / fa
+			if a == c {
+				p, q = 2*xm*s, 1-s
+			} else {
+				var r float64
+				q, r = fa/fc, fb/fc
+				p = s * (2*xm*q*(q-r) - (b-a)*(r-1))
+				q = (q - 1) * (r - 1) * (s - 1)
+			}
+
+			// Check whether in bounds.
+			if p > 0 {
+				q = -q
+			}
+
+			p = math.Abs(p)
+			if 2*p < min(3*xm*q-math.Abs(tol1*q), math.Abs(e*q)) {
+				// Accept interpolation.
+				e = d
+				d = p / q
+			} else {
+				// Interpolation failed, use bisection.
+				d = xm
+				e = d
+			}
+		} else {
+			// Bounds decreasing too slowly, use bisection.
+			d = xm
+			e = d
+		}
+
+		// Move last best guess to a.
+		a, fa = b, fb
+		// Evaluate new trial root.
+		if math.Abs(d) > tol1 {
+			b += d
+		} else {
+			b += sign(tol1, xm)
+		}
+		fb = f(b)
+	}
+	return b, errors.New("maximum iterations exceeded")
+}
+
+func sign(x, s float64) float64 {
+	if s > 0 {
+		return x
+	}
+	return -x
+}
+
+func min(x, y float64) float64 {
+	if x < y {
+		return x
+	}
+	return y
+}
--- a/root/brent_test.go
+++ b/root/brent_test.go
@@ -1,0 +1,401 @@
+package root_test
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"gonum.org/v1/exp/root"
+	"gonum.org/v1/gonum/floats/scalar"
+)
+
+var eps = math.Nextafter(1, 2) - 1
+
+func TestBrent(t *testing.T) {
+	t.Parallel()
+
+	type testcase struct {
+		name string
+		f    func(float64) float64
+		a    float64
+		b    float64
+		tol  float64
+		want float64
+	}
+	var tests []testcase
+	for _, f := range tstutilsFns {
+		tests = append(tests, testcase{name: f.name, f: f.f, a: .5, b: math.Sqrt(3), tol: 4 * eps, want: 1})
+	}
+	for _, tc := range aps {
+		tests = append(tests, testcase{name: tc.name, f: tc.f, a: tc.a, b: tc.b, tol: 4 * eps, want: tc.root})
+	}
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("%s", test.name), func(t *testing.T) {
+			root, err := root.Brent(test.f, test.a, test.b, test.tol)
+			if !scalar.EqualWithinAbsOrRel(root, test.want, 2e-12, test.tol) {
+				if err != nil {
+					t.Fatalf("%s: %s", test.name, err)
+				}
+				t.Fatalf("%s: got %f want %f", test.name, root, test.want)
+			}
+		})
+	}
+}
+
+func TestScipyIssue5557(t *testing.T) {
+	t.Parallel()
+
+	f := func(x float64) float64 {
+		if x < 0.5 {
+			return -0.1
+		}
+		return x - 0.6
+	}
+	tol := 4 * eps
+	root, err := root.Brent(f, 0., 1., tol)
+	if err != nil {
+		t.Fatalf("Brent(%f, %f, %f): %s", 0., 1., tol, err)
+	}
+	const want = 0.6
+	if !scalar.EqualWithinRel(root, want, tol) {
+		t.Fatalf("Brent(%f, %f, %f): got %f want %f", 0., 1., tol, root, want)
+	}
+}
+
+func TestScipyIssue13737(t *testing.T) {
+	t.Parallel()
+
+	// Constants based on https://github.com/scipy/scipy/blob/main/scipy/optimize/tests/test_zeros.py
+	tests := []struct {
+		a    float64
+		b    float64
+		root float64
+	}{
+		{a: -450, b: -350, root: -400},
+		{a: 350, b: 450, root: 400},
+	}
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			c := math.Exp(test.root)
+			f := func(x float64) float64 { return math.Exp(x) - c }
+			tol := eps
+			root, err := root.Brent(f, test.a, test.b, tol)
+			if err != nil {
+				t.Fatalf("Brent(%f, %f, %f): %s", test.a, test.b, tol, err)
+			}
+			if !scalar.EqualWithinRel(root, test.root, tol) {
+				t.Fatalf("Brent(%f, %f, %f): got %f want %f", test.a, test.b, tol, root, test.root)
+			}
+		})
+	}
+}
+
+func TestScipyIssue5584(t *testing.T) {
+	t.Parallel()
+
+	f := func(x float64) float64 { return 1e-200 * x }
+
+	// Report failure when signs are the same.
+	a, b := -0.5, -0.4
+	tol := eps
+	x0, err := root.Brent(f, a, b, tol)
+	if err == nil {
+		t.Fatalf("Brent(%f, %f, %f) should error", a, b, tol)
+	}
+
+	// Solve successfully when signs are different.
+	a, b = -0.5, 0.4
+	x0, err = root.Brent(f, a, b, tol)
+	if err != nil {
+		t.Fatalf("Brent(%f, %f, %f): %s", a, b, tol, err)
+	}
+	const want = 0.
+	if !scalar.EqualWithinRel(x0, want, tol) {
+		t.Fatalf("Brent(%f, %f, %f): got %f want %f", a, b, tol, x0, want)
+	}
+
+	// Solve successfully when one side is negative zero.
+	a, b = -0.5, -0.
+	x0, err = root.Brent(f, a, b, tol)
+	if err != nil {
+		t.Fatalf("Brent(%f, %f, %f): %s", a, b, tol, err)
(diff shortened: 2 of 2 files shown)


soypat:
Please consider adding the solver as a method on a type. I suggest using the following interface:
```
type RootFinder interface {
    // x0 is starting x. f is the function we seek root for.
    Root(x0 float64, f func(float64) float64) (float64, error)
}
```

You may find an adaptive newton's method solver I've used over the years here: https://github.com/soypat/glgl/blob/aae7eebe6ecdcdde211850267709f8e65e4ac642/math/md1/md1.go#L54

I've succesfully used it in many different applications thanks to it's flexible nature.

fumin:
@soypat the `RootFinder` interface you suggested seems to support only gradient based methods, which requires only a single starting x.
However, for bracket based methods including Brent, the starting object is an interval.
Moreover, among bracket based methods, there are some such as the Secant method that can work with any interval, while others such as Brent require the interval to be a "straddle" ([a, b] where f(a)*f(b) < 0).
Taking into account these possibilities, the `RootFinder` interface might look like:

```
type RootFinder interface {
    // RootGradient finds a root using a gradient based method.
    // x0 is starting x. f is the function we seek root for.
    RootGradient(x0 float64, f func(float64) float64) (float64, error)

    // RootBracket finds a root using a bracket based method.
    // [a, b] is the starting interval.
    RootBracket(a, b float64, f func(float64)float64) (float64, error)

    // RootStraddle finds a root using a straddle based method.
    // [a, b] is the starting interval.
    RootStraddle(a, b float64, f func(float64)float64) (float64, error)
}
```

I feel that downstream users might find this interface difficult to use, due to the bifurcation of parameters and their implications.

Nonetheless, thanks for sharing your implementation of Newton's method.
That should help a lot in making the package feature complete, in the sense of having at least one gradient based method.

findroot/brent_test.go
```diff
@@ -32,7 +32,7 @@ func TestBrent(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(fmt.Sprintf("%s", test.name), func(t *testing.T) {
-			root, err := root.Brent(test.f, test.a, test.b, test.tol)
+			root, err := findroot.Brent(test.f, test.a, test.b, test.tol)
```
kortschak, line 35: I think this puts the cart before the horse. I would write this as `r, err := root.Brent(test.f, test.a, test.b, test.tol)`, or more probably in a test, `got, err := root.Brent(test.f, test.a, test.b, test.tol)`.

kortschak:
I'm not convinced we need to explicitly satisfy an interface at this stage. Providing the primitive seems like a good start to me. If we need to having interface satisfaction to help that, then that should be added, and if in the future there is a need to allow polymorphic behaviour, that can be added based on these primitives, either here or by client code.

findroot/brent_test.go
fumin: Indeed, short variable names such as `r` might actually be the most common use case.
Sorry for over thinking, I have reverted the package name back to `root`, and fixed the name shadowing in the tests.

sbinet:
The package doc says it's finding roots but the package name is 'root'.
Perhaps it's a clue the package should be named 'roots'?

FYI, "my" package is also named 'roots':
https://pkg.go.dev/go-hep.org/x/exp/roots

(Apologies for the bikeshedding)

soypat:
I agree with Dan that we can worry about designing interfaces in the future. this is experimental after all.

That said I would like to note that  the NewtonRaphson root finder is also implements a bracket mode, but the bracket is part of the type :)

kortschak:
I'm not saying things shouldn't be encapsulated into types, just that the interface satisfaction can happen later.

fumin:
@sbinet Thanks for chiming in on a better package name.

After a survey of the standard library, I think our situation is more akin to the [encoding](https://pkg.go.dev/encoding@go1.24.1), [hash](https://pkg.go.dev/hash@go1.24.1), and [signal](https://pkg.go.dev/os/signal@go1.24.1) packages.
Similar to the way our package provides root-finding algorithms, these packages provide a variety of encoding and hashing methods.
On the other hand, packages that have pluralistic names such as "bytes", "slices", and "maps" tend to refer to truly enumerable collections, which I think our root package doesn't.
That said, I am happy to follow our community's decision with respect to naming.

On the topic on interfaces, perhaps we can defer designing those after we have Newton's method added into the root package?
Without a concrete implementation to tinker with, it might be a bit hard to come up with the right abstraction.

root/brent.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func sign(x, s float64) float64 {
+	if s > 0 {
+		return x
+	}
+	return -x
+}
+
+func min(x, y float64) float64 {
```
sbinet, line 104: the builtin `min` already exists.

root/brent.go
```diff
@@ -0,0 +1,109 @@
...
+			b += d
+		} else {
+			b += sign(tol1, xm)
+		}
+		fb = f(b)
+	}
+	return b, errors.New("maximum iterations exceeded")
+}
+
+func sign(x, s float64) float64 {
```
sbinet, line 97: looks like `math.Copysign` to me :}

sbinet, line 97: if I read the C and F77 version of `Section 9.3, Vol 1: Numerical Recipes in Fortran 77 2nd Ed.`, this `sign` version is bogus for `x < 0`.

we just want `math.Copysign` here.

fumin: Thanks so much for catching this, and sorry for the mistake.
I was lucky `tol1` is always positive in this algorithm, but this fix is still necessary and much better.

root/brent.go
fumin: Unfortunately, go.mod is currently set to go1.19, which doesn't support the builtin `min` function (error message in appendix A).

I updated go.mod to go1.21 locally, and all tests in the repo (not only in this root package) ran successfully.
I wonder do we want to consider bumping go.mod to above go1.21?
What's the policy in this experimental repo?
Go's x repo itself [support the previous two releases and tip](https://go.dev/wiki/X-Repositories).
Should we have similar policies, perhaps bumping go.mod is a good idea.

Appendix A: Error message when trying to use the builtin `min`
```
$ go test
# gonum.org/v1/exp/root
./brent.go:69:13: built-in min requires go1.21 or later (-lang was set to go1.19; check go.mod)
FAIL	gonum.org/v1/exp/root [build failed]
```

sbinet: it's probably fine to bump it to `go1.23` as this is what we'll do for the next Gonum release.

but others may want to chime in.

kortschak: I'm OK with this, but I think it should be in a different PR. I've sent #66 to do this, among other things (may need tuning, so currently a draft).

soypat: What about math.Min?

fumin: @soypat Thanks for the suggestion, learnt a new thing. I have applied this change.

sbinet:
BTW, I knew this discussion rang a very distant bell...
and indeed :)

https://github.com/gonum/gonum/pull/896

fumin:
> BTW, I knew this discussion rang a very distant bell... and indeed :)

Yes, I am aware of https://github.com/gonum/gonum/pull/896 , and found it a bit of a pity that it wasn't merged.
I can understand it's not easy to design a good interface for a generic root-finder that incorporates gradient-based, bracket-based, and even multivariate functions.

Hopefully, by starting from here in the exp repo, we can achieve the above goal by iterating bit by bit with more freedom to try out different designs.

fumin:
I think all comments have been addressed, I wonder could anyone give this pull request another look? Thanks.

soypat:
LGTM! We can add root functions to the package and eventually we'll find a common pattern for an interface, if that ever makes sense.

soypat (approved):

root/brent.go
```diff
@@ -0,0 +1,95 @@
...
+			e = d
+		}
+		if math.Abs(fc) < math.Abs(fb) {
+			a, fa = b, fb
+			b, fb = c, fc
+			c, fc = a, fa
+		}
+
+		// Convergence check.
+		tol1 := 2*eps*math.Abs(b) + 0.5*tol
```
kortschak, line 44: We'll see FMA/FMS on arm64 with things like this, so we probably want to coerce to `float64` precision.

(throughout)

root/brent_test.go
```diff
@@ -0,0 +1,405 @@
...
+		name string
+		f    func(float64) float64
+		a    float64
+		b    float64
+		tol  float64
+		want float64
+	}
+	var tests []testcase
+	for _, f := range tstutilsFns {
+		tests = append(tests, testcase{name: f.name, f: f.f, a: .5, b: math.Sqrt(3), tol: 4 * eps, want: 1})
```
kortschak, line 31:
```suggestion
		tests = append(tests, testcase{name: f.name, f: f.f, a: 0.5, b: math.Sqrt(3), tol: 4 * eps, want: 1})
```
(a single dot is too little syntax)

root/brent_test.go
```diff
@@ -0,0 +1,405 @@
...
+		return x - 0.6
+	}
+	tol := 4 * eps
+	got, err := root.Brent(f, 0., 1., tol)
+	if err != nil {
+		t.Fatalf("Brent(%f, %f, %f): %s", 0., 1., tol, err)
+	}
+	const want = 0.6
+	if !scalar.EqualWithinRel(got, want, tol) {
+		t.Fatalf("Brent(%f, %f, %f): got %f want %f", 0., 1., tol, got, want)
```
kortschak, line 66:
```suggestion
	got, err := root.Brent(f, 0, 1, tol)
	if err != nil {
		t.Fatalf("Brent(%f, %f, %f): %s", 0, 1, tol, err)
	}
	const want = 0.6
	if !scalar.EqualWithinRel(got, want, tol) {
		t.Fatalf("Brent(%f, %f, %f): got %f want %f", 0, 1, tol, got, want)
```
(untyped constants in a typed slot don't need extra syntax)

root/brent_test.go
```diff
@@ -0,0 +1,405 @@
...
+	if err != nil {
+		t.Fatalf("Brent(%f, %f, %f): %s", a, b, tol, err)
+	}
+	const want = 0.
+	if !scalar.EqualWithinRel(x0, want, tol) {
+		t.Fatalf("Brent(%f, %f, %f): got %f want %f", a, b, tol, x0, want)
+	}
+
+	// Solve successfully when one side is negative zero.
+	a, b = -0.5, -0.
```
kortschak, line 123: This does not work the way that you think it does; it is not possible to express a negative zero float as a literal ([spec](https://go.dev/ref/spec#Constants), "Consequently, there are no constants denoting the IEEE 754 negative zero, infinity, and not-a-number values." https://go.dev/play/p/8VxCUsm3Uum). You can use [`math.Copysign(0, -1)
`](https://go.dev/play/p/AOWBduVzEce)

root/brent_test.go
```diff
@@ -0,0 +1,405 @@
...
+		case x < 1:
+			return -1 + 0.1*x
+		default:
+			return 0
+		}
+	}},
+	// f5 is a hyperbola with a pole at x=1, but pole replaced with 0. Not continuous at root.
+	{name: "f5", f: func(x float64) float64 {
+		if x != 1 {
+			return 1. / (1 - x)
```
kortschak, line 158:
```suggestion
			return 1 / (1 - x)
```

root/brent_test.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+// aps11 is a rational function with a zero at x=1/n and a pole at x=0.
+func aps11(x, n float64) float64 {
+	return (n*x - 1) / ((n - 1) * x)
+}
+
+// aps12 is nth root of x, with a zero at x=n.
+func aps12(x, n float64) float64 {
+	return math.Pow(x, 1.0/n) - math.Pow(n, 1.0/n)
```
kortschak, line 225:
```suggestion
	return math.Pow(x, 1/n) - math.Pow(n, 1/n)
```

root/brent_test.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+// aps14 returns 0 for negative x-values, trigonometric+linear for x positive.
+func aps14(x, n float64) float64 {
+	if x <= 0 {
+		return -n / 20
+	}
+	return n / 20.0 * (x/1.5 + math.Sin(x) - 1)
```
kortschak, line 233:
```suggestion
	return n / 20 * (x/1.5 + math.Sin(x) - 1)
```

kortschak (commented):
This is outside my area of expertise, so leaving to others, but for light review.

root/brent_test.go
fumin: Thanks for the suggestion, I have applied them to the function call to `Brent`.
However, for the call to `Fatalf`, string formatting for %f seems to insist on the additional dot, thus I have kept it.

root/brent_test.go
fumin: Thanks for this interesting educating advice. I have applied this change.

root/brent_test.go
kortschak: Good point; the reason for that is that the args are `any`, so the default type is `int` when there is no syntax to say otherwise. Please add the following zero then; `, 0.0, 1.0`.

fumin: Yes, I have applied this change in a hopefully more principled way, by defining the parameters as float64 variables, and inserting them throughout.

fumin:
@sbinet I wonder if you can take another look at this change? All previous comments have been addressed.

soypat:
Is there anything blocking the merge?

sbinet (commented):
apologies for the belated answer.

here are a couple of comments.
@kortschak (and others) may want to chime in.

fumin:
@sbinet thanks for the review, all comments have been addressed and incorporated.

root/brent.go
```diff
@@ -0,0 +1,101 @@
...
+package root
+
+import (
+	"errors"
+	"math"
+)
+
+var (
+	ErrInterval        = errors.New("root: invalid root bracket")
+	ErrMaxIterExceeded = errors.New("maximum iterations exceeded")
```
sbinet, line 14:
```suggestion
	ErrMaxIter = errors.New("root: maximum iterations exceeded")
```

root/brent_test.go
```diff
@@ -0,0 +1,405 @@
...
+
+	f := func(x float64) float64 { return 1e-200 * x }
+
+	// Report failure when signs are the same.
+	a, b := -0.5, -0.4
+	tol := eps
+	x0, err := root.Brent(f, a, b, tol)
+	if err != root.ErrInterval {
+		t.Fatalf("Brent(%f, %f, %f) should error", a, b, tol)
+	}
```
sbinet, line 110:
```suggestion
	if got, want := err, root.ErrInterval; !errors.Is(got, want) {
		t.Fatalf("Brent(%f, %f, %f) should error: got=%q, want=%q", a, b, tol, got, want)
	}
```

sbinet (commented):
last nitpicks from my end.

fumin:
@sbinet thanks for the comments, they have been incorporated.

sbinet (dismissed):
LGTM

Thanks for sticking with us.
I'll wait for a couple of days to let others chime in.

(and feel free to ping us if we let the ball fall on the floor)

root/brent_test.go
```diff
@@ -0,0 +1,406 @@
...
+		tol  float64
+		want float64
+	}
+	var tests []testcase
+	for _, f := range tstutilsFns {
+		tests = append(tests, testcase{name: f.name, f: f.f, a: 0.5, b: math.Sqrt(3), tol: 4 * eps, want: 1})
+	}
+	for _, tc := range aps {
+		tests = append(tests, testcase{name: tc.name, f: tc.f, a: tc.a, b: tc.b, tol: 4 * eps, want: tc.root})
+	}
```
kortschak, line 36: I'm not a fan of this construction. Let's move all of the cases in `aps` and `tstutilsFns` into a single test package var statically rather than constructing this slice in the test. Indirection is something that we generally avoid, particularly in test code.
```
var brentTests = []struct {
	name string
	f    func(float64) float64
	a    float64
	b    float64
	tol  float64
	want float64
} {
	…
}
```

kortschak (changes requested):

root/brent_test.go
fumin: Indeed, thanks for the suggestion. I have done away with the scattered `aps` and `tsutilsFns` indirections, and consolidated them into the package var `brentTests`.

sbinet (approved):
Still LGTM.

@fumin could you send another PR against the main gonum/gonum repository, adding yourself to the `AUTHORS` and/or `CONTRIBUTORS` files ?
(then, I'll merge this PR)

thanks again.

fumin:
@sbinet yes of course, I think https://github.com/gonum/gonum/pull/2028 does this. Thanks for the review.

PR #67, opened 2025-03-10
Plot bump

sbinet:
Please take a look.


kortschak:
The line appearance in the first four are correct; I never noticed that they were missing.

The shift in the text is likely due to text metric changes between the two and I'm fine with those.

sbinet:
OK. I'll make this a real PR when you'll merge yours :)

kortschak (approved):

PR #71, opened 2025-03-15
migrate to math/rand/v2

### LINKED ISSUE #69: all: migrate to math/rand/v2
when Gonum@v0.16.x is out (and has the support for `math/rand/v2` exported), we should migrate `linsolve` and `plotter/rings` to `math/rand/v2`.

sbinet:
Please take a look.

Fixes #69.


sbinet:
restacked on top of https://github.com/gonum/exp/pull/73

plotter/rings/bezier.go
```diff
@@ -72,6 +71,17 @@ type Bezier struct {
 	// If nil, these values are not used.
 	Crest  *FactorDist
 	Purity *FactorDist
+
+	// Rand is a source of random numbers.
+	// If nil, the global one is used.
+	Rand *rand.Rand
```
kortschak: Can we use a `rand.Source` here instead?

sbinet: right. (like for `gonum/stat/distuv` and friends.)

done.

sbinet:
PTAL

PR #72, opened 2025-06-08
root: add FindBracketMono for finding brackets that hug a function's root

### LINKED ISSUE #70: proposal: root: add FindBracketMono to find a suitable bracket for use in bracket based root finding algorithms
### Background
We will soon have [bracket based root finding algorithms](https://github.com/gonum/exp/pull/65) in the root package.
In general, users are expected to leverage their knowledge about the function to come up with a bracket interval [a, b] that hugs the root, i.e. f(a)f(b) < 0.
However, in the special case of monotonic functions, we can actually help users identify such a bracket automatically.
This will make using our root finding algorithms much more convenient in statistical analysis, where the Cumulative function is a monotonically increasing function.

### Proposal
I propose the follow signature:

```
// FindBracketMono finds a bracket interval [a, b] where f(a)f(b) < 0.
// f must be a monotonically increasing function.
func FindBracketMono(f func(float64) float64, guess float64) (float64, float64)
```

The algorithm behind  `FindBracketMono` is that of the Boost C++ library's [bracketing routine](https://live.boost.org/doc/libs/1_87_0/libs/math/doc/html/math_toolkit/roots_noderiv/bracket_solve.html).
What this algorithm does is leveraging the fact that for monotonic functions, we always know whether the root is towards 0 or Inf, relative to the current guess.

### Potential impact of proposal

This function will be placed in the new root package. I will do the work.

fumin:
fixes https://github.com/gonum/exp/issues/70

Please take a look.

### DIFF
--- a/root/findbracket.go
+++ b/root/findbracket.go
@@ -8,6 +8,7 @@ import "math"
 
 // FindBracketMono finds a bracket interval [a, b] where f(a)f(b) < 0.
 // f must be a monotonically increasing function.
+// guess is the initial guess of the bracket search.
 func FindBracketMono(f func(float64) float64, guess float64) (a, b float64) {
 	// Make sure initial guess has the same sign as the root.
 	f0 := f(0)
@@ -43,9 +44,16 @@ func FindBracketMono(f func(float64) float64, guess float64) (a, b float64) {
 	// If unable to cross x-axis, return the largest possible bracket.
 	if !crossed {
 		if r > 1 {
-			return a, math.Inf(int(math.Copysign(1, b)))
+			b = math.Inf(int(math.Copysign(1, b)))
+		} else {
+			b = 0
 		}
-		return a, 0
 	}
+
+	// Ensure a <= b
+	if a > b {
+		a, b = b, a
+	}
+
 	return a, b
 }
--- a/root/findbracket.go
+++ b/root/findbracket.go
@@ -1,0 +1,46 @@
+package root
+
+import "math"
+
+// FindBracketMono finds a bracket interval [a, b] where f(a)f(b) < 0.
+// f must be a monotonically increasing function.
+func FindBracketMono(f func(float64) float64, guess float64) (float64, float64) {
+	// Make sure initial guess has the same sign as the root.
+	f0 := f(0)
+	if (guess < 0 && f0 < 0) || (guess > 0 && f0 > 0) {
+		guess *= -1
+	}
+
+	// r is the rate in which we adjust the interval.
+	var r float64
+	a, fa := guess, f(guess)
+	if (a > 0) == (fa < 0) {
+		r = 2
+	} else {
+		r = 1. / 2
+	}
+
+	// Expand bracket until x-axis is crossed.
+	crossed := false
+	b := a * r
+	fb := f(b)
+	for range 200 {
+		if math.Signbit(fa) != math.Signbit(fb) || fa == 0 || fb == 0 {
+			crossed = true
+			break
+		}
+		a, fa = b, fb
+		b *= r
+		fb = f(b)
+	}
+	// If unable to cross x-axis, return the largest possible bracket.
+	if !crossed {
+		if r > 1 {
+			return a, math.Inf(int(math.Copysign(1, b)))
+		} else {
+			return a, 0
+		}
+	}
+
+	return a, b
+}
--- a/root/findbracket.go
+++ b/root/findbracket.go
@@ -1,10 +1,14 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package root
 
 import "math"
 
 // FindBracketMono finds a bracket interval [a, b] where f(a)f(b) < 0.
 // f must be a monotonically increasing function.
-func FindBracketMono(f func(float64) float64, guess float64) (float64, float64) {
+func FindBracketMono(f func(float64) float64, guess float64) (a, b float64) {
 	// Make sure initial guess has the same sign as the root.
 	f0 := f(0)
 	if (guess < 0 && f0 < 0) || (guess > 0 && f0 > 0) {
@@ -17,14 +21,16 @@ func FindBracketMono(f func(float64) float64, guess float64) (float64, float64) 
 	if (a > 0) == (fa < 0) {
 		r = 2
 	} else {
-		r = 1. / 2
+		r = 0.5
 	}
 
 	// Expand bracket until x-axis is crossed.
+	// maxiter value is based on https://github.com/boostorg/math/blob/boost-1.88.0/include/boost/math/policies/policy.hpp#L130
+	const maxiter = 200
 	crossed := false
-	b := a * r
+	b = a * r
 	fb := f(b)
-	for range 200 {
+	for range maxiter {
 		if math.Signbit(fa) != math.Signbit(fb) || fa == 0 || fb == 0 {
 			crossed = true
 			break
@@ -37,10 +43,8 @@ func FindBracketMono(f func(float64) float64, guess float64) (float64, float64) 
 	if !crossed {
 		if r > 1 {
 			return a, math.Inf(int(math.Copysign(1, b)))
-		} else {
-			return a, 0
 		}
+		return a, 0
 	}
-
 	return a, b
 }
--- a/root/findbracket.go
+++ b/root/findbracket.go
@@ -17,7 +17,8 @@ func FindBracketMono(f func(float64) float64, guess float64) (a, b float64) {
 
 	// r is the rate in which we adjust the interval.
 	var r float64
-	a, fa := guess, f(guess)
+	a = guess
+	fa := f(a)
 	if (a > 0) == (fa < 0) {
 		r = 2
 	} else {
--- a/root/findbracket_test.go
+++ b/root/findbracket_test.go
@@ -1,0 +1,93 @@
+package root_test
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"gonum.org/v1/exp/root"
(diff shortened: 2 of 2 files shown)


root/findbracket.go
```diff
@@ -0,0 +1,46 @@
+package root
```
kortschak, line 5: Please add the copyright header.

root/findbracket.go
```diff
@@ -0,0 +1,46 @@
...
+		guess *= -1
+	}
+
+	// r is the rate in which we adjust the interval.
+	var r float64
+	a, fa := guess, f(guess)
+	if (a > 0) == (fa < 0) {
+		r = 2
+	} else {
+		r = 1. / 2
```
kortschak:
```suggestion
		r = 0.5
```

root/findbracket.go
```diff
@@ -0,0 +1,46 @@
...
+func FindBracketMono(f func(float64) float64, guess float64) (float64, float64) {
+	// Make sure initial guess has the same sign as the root.
+	f0 := f(0)
+	if (guess < 0 && f0 < 0) || (guess > 0 && f0 > 0) {
+		guess *= -1
+	}
+
+	// r is the rate in which we adjust the interval.
+	var r float64
+	a, fa := guess, f(guess)
```
kortschak: Is there a reason this is on a single line while below the `f(b)` evaluation is not?

root/findbracket.go
```diff
@@ -0,0 +1,46 @@
...
+		b *= r
+		fb = f(b)
+	}
+	// If unable to cross x-axis, return the largest possible bracket.
+	if !crossed {
+		if r > 1 {
+			return a, math.Inf(int(math.Copysign(1, b)))
+		} else {
+			return a, 0
+		}
```
kortschak:
```suggestion
		if r > 1 {
			return a, math.Inf(b)
		}
		return a, 0
```

root/findbracket_test.go
```diff
@@ -0,0 +1,93 @@
+package root_test
```
kortschak, line 5: Copyright header.

root/findbracket_test.go
```diff
@@ -0,0 +1,93 @@
...
+			a, b := root.FindBracketMono(test.f, test.guess)
+			fa, fb := test.f(a), test.f(b)
+			if fa*fb > 0 {
+				t.Fatalf("%s: invalid bracket (%f, %f)", test.name, fa, fb)
+			}
+		})
+	}
+}
+
+func TestFindBracketMonoSpecialCases(t *testing.T) {
```
kortschak: Can this be table driven?

root/findbracket.go
```diff
@@ -0,0 +1,46 @@
...
+		r = 2
+	} else {
+		r = 1. / 2
+	}
+
+	// Expand bracket until x-axis is crossed.
+	crossed := false
+	b := a * r
+	fb := f(b)
+	for range 200 {
```
kortschak: Why 200?

root/findbracket.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	// If unable to cross x-axis, return the largest possible bracket.
+	if !crossed {
+		if r > 1 {
+			return a, math.Inf(int(math.Copysign(1, b)))
+		} else {
+			return a, 0
+		}
+	}
+
```
kortschak, line 52:
```suggestion
```

root/findbracket.go
```diff
@@ -0,0 +1,46 @@
+package root
+
+import "math"
+
+// FindBracketMono finds a bracket interval [a, b] where f(a)f(b) < 0.
+// f must be a monotonically increasing function.
+func FindBracketMono(f func(float64) float64, guess float64) (float64, float64) {
```
kortschak:
```suggestion
func FindBracketMono(f func(float64) float64, guess float64) (a, b float64) {
```

root/findbracket.go
fumin: My preference is for all lines to be single as it is more concise.
However, the below `f(b)` evaluations have to be done *after* `b` is modified.

In my opinion, the following single line version:

```b, fb = b*r, f(b)```

invites too much ambiguity, and relies too much on familiarity of subtle operation precedence rules.
Therefore, I opted for the one line one statement version for the `f(b)` evaluations.

That said, I am eager to learn others opinions on this topic, and perhaps if there are coding conventions within gonum to follow them.

File: root/findbracket.go (hunk 4 above)
Comment (fumin) [reply root/findbracket.go]: I need some help in simplifying the expression `math.Inf(int(math.Copysign(1, b)))`.

The expression `math.Inf(b)` is ideal, but alas `math.Inf` requires its argument to be an `int`, whereas `b` is a float64.
`math.Inf(int(b))` does not work, since for `b=-0.5`, we want `-Inf` but due to the rules of `math.Inf` itself and of the conversion from int to float64, becomes `+Inf`.
Neither does wrapping with `math.Ceil` nor `math.Round` do the trick.

File: root/findbracket_test.go (hunk 6 above)
Comment (fumin) [reply root/findbracket_test.go]: I tried my best to make this table driven in my initial code, but unfortunately was not able to make it.
The difficulty lies in the fact that the checks are not simple equal or roughly equal comparisons.
Perhaps it could be done by designing the test cases as:

```
cases := []struct{
  ...
  criteria func(a, b float64) bool
}{
  {
    criteria: func(a, b float64)bool{ return a > 0 && math.IsInf(b, 1) },
  },
  ...
}

for _, tc := range cases {
  t.Run(tc.name, func(t *testing.T){
    a, b := root.FindBracketMono(...)
    if !tc.criteria(a, b) {
      t.Fatalf("...")
    }
  })
}
```

Perhaps the above is actually not bad an idea?

File: root/findbracket.go (hunk 7 above)
Comment (fumin) [reply root/findbracket.go]: This `maxiter` value is based on https://github.com/boostorg/math/blob/boost-1.88.0/include/boost/math/policies/policy.hpp#L130 .
I have added this reference as a comment in the latest commit.

File: root/findbracket.go (hunk 4 above)
Comment (kortschak) [reply root/findbracket.go]: No, you are correct.

File: root/findbracket.go (hunk 3 above)
Comment (kortschak) [reply root/findbracket.go]: My preference is to not use multiple assignments unless they are required for the logic of the program.

Comment (fumin) [reply root/findbracket.go]: Makes sense, thanks for the suggestion. I have updated this line to single assignments.

**Comment by fumin** [comment]
Thanks for the feedback, I have incorporated all of them in the latest code.

File: root/findbracket.go [hunk 10]
```diff
@@ -0,0 +1,51 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package root
+
+import "math"
+
+// FindBracketMono finds a bracket interval [a, b] where f(a)f(b) < 0.
+// f must be a monotonically increasing function.
+func FindBracketMono(f func(float64) float64, guess float64) (a, b float64) {
```
vladimir-ch, line 12: Is there any promise that a <= b?

root/findbracket.go
```diff
@@ -0,0 +1,51 @@
+// Copyright ©2025 The Gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package root
+
+import "math"
+
+// FindBracketMono finds a bracket interval [a, b] where f(a)f(b) < 0.
+// f must be a monotonically increasing function.
```
vladimir-ch, line 10: This should have a comment about what `guess` is.

root/findbracket_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"gonum.org/v1/exp/root"
+)
+
+var findBracketMonoTests = []struct {
```
vladimir-ch, line 14: I'd have a single test function and a single table with the `criteria` function renamed as `validBracket` (for example) and if `validBracket` is nil, then use something like `func(f func(float64) float64, a, b float64) bool { return f(a)*f(b) <= 0 }`.

root/findbracket_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Based on https://github.com/boostorg/math/blob/boost-1.88.0/test/test_root_finding_concepts.cpp
+	{name: "f1", f: func(x float64) float64 { return x*x*x - 27 }, guess: 27},
+}
+
+func TestFindBracketMono(t *testing.T) {
+	t.Parallel()
+
+	for _, test := range findBracketMonoTests {
+		t.Run(fmt.Sprintf("%s", test.name), func(t *testing.T) {
```
vladimir-ch: No need for `fmt.Sprintf`.

root/findbracket_test.go
```diff
@@ -0,0 +1,86 @@
...
+			a, b := root.FindBracketMono(test.f, test.guess)
+			fa, fb := test.f(a), test.f(b)
+			if fa*fb > 0 {
+				t.Fatalf("%s: invalid bracket (%f, %f)", test.name, fa, fb)
```
vladimir-ch: - It doesn't make any difference here but I'd use `Errorf` instead.
- The log message is misleading, it says "bracket" (expectation a and b) but prints f(a) and f(b).

root/findbracket.go
fumin: There wasn't, but the expression `[a, b]` in the function doc implies that there should be.
I have updated the code to ensure `a <= b`.
Thanks for this very good point.

root/findbracket_test.go
fumin: Good point about the fact that "bracket" should refer to `(a, b)` not `(f(a), f(b))`.
I have updated the code accordingly.

fumin:
@vladimir-ch thanks for the review, all suggestions have been applied in the latest code.

